PR #224, opened 2023-04-11
update account entity

OmolemoBlessingLethuloe:
Getting error because account entity was conflicting with account on shesha core


OmolemoBlessingLethuloe:
image

ktsapo (approved):

PR #226, opened 2023-04-11
Phumudzo/fix code editor

nthangeniphumudzo:
Hi Ivan, this fix is related to [this card](https://dev.azure.com/boxfusion/PD-Dep/_workitems/edit/46811/). I have figured that the doublication of characters only happen when the "Autocompletion" properties are true on the 'CodeEditor' component so I have assigned them to false. Please approve my request if this doesn't cause any other discrepancies in the long run. I have also pushed minor style change for the imageAnnotation component.


IvanIlyichev:
Hi @nthangeniphumudzo. This bug requires a proper fix. You only disabled the autocompletion and did just in few places. We should fix the autocompletion instead of disabling

IvanIlyichev:
old PR, code editor has been replaced with the monaco editor

PR #228, opened 2023-04-12
updated discriminator column to have a default value

yolandmagxagxa:


IvanIlyichev (changes requested):
In this case you should edit existing migration M20230411175800 because it will fail if rows already exists in the table and your new migration will not be applied at all. Please modify M20230411175800 and remove M20230412095600. Note: empty string is not a valid value of the discriminator, existing rows fill be filtered out by the NH, you need to specify a correct value. You can check current value of the discriminator generated by the conventions using NHibernate endpoint

IIlyichev (approved):

PR #239, opened 2023-04-20
get bank with members custom API

OmolemoBlessingLethuloe:


shesha-functional-tests/backend/src/Boxfusion.SheshaFunctionalTests.Common.Application/Services/BankMemberAppService.cs
```diff
@@ -27,13 +29,30 @@ public BankMemberAppService(IRepository<Bank, Guid> bankRepo)
 
             input.Members.ForEach(async member =>
             {
-                await SaveOrUpdateEntityAsync<Member>(member, async item =>
+                using (var uow = UnitOfWorkManager.Begin())
```
ktsapo, line 32: Why is the unit of work opened here?

PR #265, opened 2023-05-16
Adding entities and app services to test data table and map component

OmolemoBlessingLethuloe:


shesha-functional-tests/backend/src/Boxfusion.SheshaFunctionalTests.Common.Application/Services/SchoolAppService.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// <returns></returns>
+        /// <exception cref="AbpValidationException"></exception>
+        public async Task<DynamicDto<School, Guid>> CreateSchoolCustom (DynamicDto<School, Guid> input)
+        {
+            var school = await SaveOrUpdateEntityAsync<School>(null, async item =>
+            {
+                var validationResults = new List<ValidationResult>();
+
+                var result = await MapJObjectToEntityAsync<School, Guid>(input._jObject, item, validationResults);
+                if (!result) throw new AbpValidationException("Please correct the errors and try again", validationResults);
```
ktsapo, line 24: Wouldn't it make sense to check for the validation results here instead of checking the result?

PR #317, opened 2023-06-12
made changes on FullAuditedEntityWithExternalSync changing string len…

jonathan-boxfusion:
…ght on properties


IvanIlyichev (approved):
@jonathan-boxfusion don't forget to update columns in the DB

PR #347, opened 2023-06-28
feat :change header size

pholoshos:


Lukeybooi (commented):
Please use the style sheet instead of inline styling

Lukeybooi (approved):

PR #427, opened 2023-09-21
Added properties on account

MachoeneMoichela:
Added properties on the Account. entity


IvanIlyichev:
Properties have been added as part of another PR

PR #443, opened 2023-10-18
Fix/le/date picker

Lukeybooi:
- Fix date picker functionality on the data table


IvanIlyichev (changes requested):
Please provide comments in the code. Defaults values require a refactoring, currently they are handled by a wrong way. Your comments will help to keep required functionality during this refactoring, otherwise your changes will be broken

IvanIlyichev (approved):

PR #452, opened 2023-11-01
User/machoene/fix added dsicriminator on mobiledevice

MachoeneMoichela:


IvanIlyichev:
The same changes have been made as part of another PR

PR #482, opened 2023-11-10
F/blobstorage configs

azarielmochusi:


IvanIlyichev (commented):
Hi @azarielmochusi. Have you tested it when **CloudStorage** is missing in the appsettings?
What's the purpose of the **View** Action? It's a copy of the **Download** but without call of **MarkDownloadedAsync**. Note: content-disposition header may not work, it depends on the Access-Control-Expose-Headers. To return a filename you can use ShaFileStreamResult/ShaFileContentResult

PR #487, opened 2023-11-13
Feat/le/quick form edit

Lukeybooi:
- Feat: added dynamic modal on form info


IvanIlyichev (changes requested):
The **FormInfoContentProvider** adds unneeded complexity and breaks the form scope. You use a single instance of the **FormInfoContentProvider** but the application can render multiple configurable forms on the same page. The functionality implemented in this provider should be a part of the FormProvider or ConfigurableForm, it this case each ConfigurableForm will have it's own instance of flags and handlers

IvanIlyichev (changes requested):
@Lukeybooi Build failed and you still use effects for control flows, the code requires refactoring

IvanIlyichev:
Implemented and merged using another pull request

PR #585, opened 2023-11-21
fix (Notifications) : fixed bug where Recipient is not stored in DB.

Ern7:
Fixed bug where Recipient is not stored in DB during the sending of an sms, email or push notification.

### DIFF
--- a/shesha-core/src/Shesha.Application/NotificationMessages/NotificationMessageAppService.cs
+++ b/shesha-core/src/Shesha.Application/NotificationMessages/NotificationMessageAppService.cs
@@ -1,5 +1,6 @@
 ﻿using Abp.Domain.Repositories;
 using Shesha.Domain;
+using Shesha.DynamicEntities.Dtos;
 using Shesha.NotificationMessages.Dto;
 using Shesha.Notifications;
 using System;
@@ -32,5 +33,19 @@ public class NotificationMessageAppService : SheshaCrudServiceBase<NotificationM
         await _distributer.ResendMessageAsync(dto);
 
         return true;
+    }
+
+    public async Task<DynamicDto<NotificationMessage, Guid>> MarkAsReadAsync(Guid id)
+    {
+        if (id == Guid.Empty)
+            throw new ArgumentNullException(nameof(id));
+
+        var entity = await SaveOrUpdateEntityAsync<NotificationMessage>(id, item =>
+        {
+            item.Opened = true;
+            item.LastOpened = DateTime.UtcNow;
+        });
+
+        return await MapToDynamicDtoAsync<NotificationMessage, Guid>(entity);
     }
 }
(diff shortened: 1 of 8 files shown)


shesha-core/src/Shesha.Application/NotificationMessages/NotificationMessageAppService.cs
```diff
@@ -33,4 +34,18 @@ public async Task<bool> Resend(Guid id)
 
         return true;
     }
+
+    public async Task<DynamicDto<NotificationMessage, Guid>> MarkAsReadAsync(Guid id)
+    {
+        if (id == Guid.Empty)
+            throw new ArgumentNullException(nameof(id));
+
+        var entity = await SaveOrUpdateEntityAsync<NotificationMessage>(id, item =>
```
ihouvet, line 43: What should happen if the message was already marked as opened previously? Should it overwrite the timestamp? Probably not I think.

shesha-core/src/Shesha.Application/Notifications/INotificationAppService.cs
```diff
@@ -45,9 +46,10 @@ public interface INotificationAppService: IDynamicCrudAppService<Notification, D
         /// <param name="emailAddress">Recipient email address</param>
         /// <param name="attachments">Notification attachments</param>
         /// <param name="sourceEntity">Optional parameter. If notification is an Entity level notification, specifies the entity the notification relates to.</param>
+        /// <param name="recipient">Receiptient of the notification</param>
         /// <param name="cc"></param>
         /// <returns></returns>
-        Task PublishEmailNotificationAsync<TData>(Guid templateId, TData data, string emailAddress, List<NotificationAttachmentDto> attachments, GenericEntityReference sourceEntity = null, string cc = "") where TData : NotificationData;
+        Task PublishEmailNotificationAsync<TData>(Guid templateId, TData data, string emailAddress, List<NotificationAttachmentDto> attachments, GenericEntityReference sourceEntity = null, Person recipient = null, string cc = "") where TData : NotificationData;
```
ihouvet: We are creating a potential contradiction here whereby the email address provided may not necessarily be the email address of the Person provided.
I think it would be safer to have a different overload where you only pass in the Person as a mandatory parameter and it pulls out the email address from the Person entity. This eliminates the possibility of the contradiction.

shesha-core/src/Shesha.Application/Notifications/INotificationAppService.cs
```diff
@@ -56,18 +58,20 @@ public interface INotificationAppService: IDynamicCrudAppService<Notification, D
         /// <param name="data">Data that is used to fill template</param>
         /// <param name="mobileNo">Recipient mobile number</param>
         /// <param name="sourceEntity">Optional parameter. If notification is an Entity level notification, specifies the entity the notification relates to.</param>
+        /// <param name="recipient">Recipient of the notification</param>
         /// <returns></returns>
-        Task PublishSmsNotificationAsync<TData>(string notificationName, TData data, string mobileNo, GenericEntityReference sourceEntity = null) where TData : NotificationData;
-
+        Task PublishSmsNotificationAsync<TData>(string notificationName, TData data, string mobileNo, GenericEntityReference sourceEntity = null, Person recipient = null) where TData : NotificationData;
```
ihouvet: Same comment as above applies.

ihouvet (changes requested):
Main issues is that with the changes to the SendEmail signature made you are creating a potential contradiction here whereby the email address provided may not necessarily be the email address of the Person provided.
I think it would be safer to have a different overload where you only pass in the Person as a mandatory parameter and it pulls out the email address from the Person entity. This eliminates the possibility of the contradiction.

ihouvet (approved):

PR #772, opened 2023-12-14
Updated migration to include nullable on columns

MaseruKganyago:


IvanIlyichev (changes requested):
Hi @MaseruKganyago. Please make the following changes:
1. revert change in the connection string
2. Add a new migration for altering the same columns, it's required for databases which already updated using migration M20231204141400. Text of the migration:
```
Alter.Column("TargetNotifiers").OnTable("AbpUserNotifications").AsString(1024).Nullable();
Alter.Column("TargetNotifiers").OnTable("AbpNotifications").AsString(1024).Nullable();
```

IvanIlyichev:
Hi @MaseruKganyago. Your new migration creates a column again, but it should alter. Please fix

PR #1011, opened 2024-01-31
Updated starter template

Lukeybooi:
- Starter Template Modified
- Using Latest Ant & Next JS


IvanIlyichev (commented):
Hi @Lukeybooi. Starter template was updated earlier and it already use latest versions of Antd and NextJs, your PR contains a lot of files which were removed. Check the latest template

Issue #861, opened 2024-01-11
Execute script event handler triggers


hlayisanimazuze:
Expose functions that make possible for us to handle success and failure from the execute script function. This could be useful in cases where would like to add additional conditions to the events such as rendering a dialog based on if a table has data,
e.g.```
 ```
if (globalState.dataTableContext.tableData.length) {
   handleSuccess()
} else {
  handleFail()
}
```
```
 or if you want to handle success after certain amount of time has elapsed e.g. if you have a `setTimeOut` function

Example: https://function-adminportal-test.shesha.dev/dynamic/Boxfusion.SheshaFunctionalTests.Common/Books-Table
1. First step you want to execute a script and when the script is done executing
2. Handle success and navigate to the details view of the record

hlayisanimazuze:
This item was not implemented as the current changes adequately address the requirements for scenario 1. For scenario 2, a separate card has been created to fulfill the requirement: #1068

PR #1035, opened 2024-02-06
ExecuteScriptAsync added on Action configuration

Lukeybooi:
- Exposed `executeScriptAsync` function for async expressions
- [Link](https://github.com/shesha-io/shesha-framework/issues/861)


IvanIlyichev (commented):
Hi @Lukeybooi. I've checked the code and the issue you mentioned. Please clarify how your changes cover the requirements. The code of your new action is identical to the existing `Execute script` action but it doesn't wrap a result into promise. It means that you can't guarantee correct handling of the result in the handlers chain

Issue #758, opened 2023-12-13
Freeze Panes for Better Table Navigation


hlayisanimazuze:
**Description**
As a User working with a lot of data, I want the ability to freeze panes in a table so that I can improve the visibility and navigation of my data.

**Acceptance Criteria**
1. Provide a checkbox for the following
    - Freeze headings
    - Freeze columns (This should freeze the first three columns)
    - if check it should enable freezing of panes
2. Allow some flexibility to freeze both columns and headings simultaneously
3. There should be a visual cue or indicator, such as a line or shading, to clearly demarcate the frozen panes. This will help me easily identify the frozen area and distinguish it from the scrollable content.

### Related Item
- #581

PR #1057, opened 2024-02-12
Phumudzo/en/758

nthangeniphumudzo:
Hi @IvanIlyichev please kindly review my PR related to[ this issue](https://github.com/shesha-io/shesha-framework/issues/758) and feedback I believe you are the one more knowledgable on the index table issues. Thanks


IvanIlyichev (changes requested):
Hi @nthangeniphumudzo. The code is not compilable. Make sure that `npm run build` gives no errors

shesha-reactjs/src/components/reactTable/index.tsx
```diff
@@ -378,24 +388,53 @@ export const ReactTable: FC<IReactTableProps> = ({
...
                 <div
                   key={key}
                   {...headerGroupProps}
                   className={classNames(styles.tr, styles.trHead)}
+                  style={{ ...fixedHeadersStyle, display: 'flex' }} // Make the header row sticky
                 >
-                  {headerGroup?.headers?.map((column) => {
-                    const { key, ...headerProps } = column.getHeaderProps(column.getSortByToggleProps());
+                  {headerGroup?.headers?.map((column, index) => {
+                    const isFixed = (column as any)?.isFixed;
```
IvanIlyichev: @nthangeniphumudzo avoid cast to `any`, we should use types/interfaces, otherwise your code may be broken later

shesha-reactjs/src/components/dataTable/index.tsx
```diff
@@ -318,6 +335,11 @@ export const DataTable: FC<Partial<IIndexTableProps>> = ({
             : undefined;
         const width = strictWidth ?? columnItem.width;
 
+        const isFixed =
+          (['action', 'crud-operations'].includes(columnItem.columnType) && hasFixedColumns) || columnItem?.isFixed;
```
IvanIlyichev: @nthangeniphumudzo action and crud-operations columns must work according to their settings without any non-trivial logic. It's not transparent why these columns are fixed when you have any other fixed columns

IvanIlyichev (commented):
Hi @nthangeniphumudzo. Check my inline comments. Also I think it would be good if we provide left/right options in the initial implementation to avoid rework in future.  Check this example https://codesandbox.io/p/sandbox/react-table-fixed-columns-official-kowzlm5jp7

IvanIlyichev (approved):

Issue #1066, opened 2024-02-14
Query variable is logging an empty object and also expose the variable


hlayisanimazuze:
Example: When you console `query `variable, it logs an empty object. I've consoled query onData Loaded

**Steps to reproduce the behavior:**
1. Go to 'https://function-adminportal-test.shesha.dev/dynamic/TestModule/Subformtest-details?id=8b2ca05e-f2bd-4267-afcb-09a186ff87cb'
2. Open your console page
3. Reload the page
4. Check what's being logged
5. Actual Results: The query displays an empty object

**Expected behavior**
- Expose the query variable on places where it is required
- When consoled, the query should be logged

<img width="952" alt="image" src="https://github.com/shesha-io/shesha-framework/assets/151041759/c661e12f-66a7-4481-a116-73ee9026167f">

PR #1078, opened 2024-02-18
fix: query variable display

Lukeybooi:
# Query Variable Not Logging Fix
- Allow exposed variable to display correct values
- [Link](https://github.com/shesha-io/shesha-framework/issues/1066)


IvanIlyichev (changes requested):
Hi @Lukeybooi. The code will throw an error if url is null or undefined

IvanIlyichev (approved):

Issue #1059, opened 2024-02-12
Capitalize the 'refListName' that shows when you hover on referenceListStatus


hlayisanimazuze:
- The reference list status component's displayName is in upperCase, however when you hover the font changes to lowerCase

**Expected Results:** When you hover, the font should always be in upperCase

<img width="118" alt="image" src="https://github.com/shesha-io/shesha-framework/assets/151041759/058141e4-55cd-4fb3-b689-b4b721b935e0">

PR #1111, opened 2024-02-26
Phumudzo/f/1059

nthangeniphumudzo:
Hi @AlexStepantsov  please assist with the PR addressing [this issue](https://github.com/shesha-io/shesha-framework/issues/1059).Thanks


nthangeniphumudzo:
Hi @AlexStepantsov . It seems like the Tooltip component creates a new DOM node for the tooltip content, which is not a direct child of the component where the Tooltip is used.Hence the introduction of inline style  to style the tooltip's content.

AlexStepantsov (approved):

PR #1130, opened 2024-03-04
fix: adding web config files on starter

OmolemoBlessingLethuloe:


IvanIlyichev (commented):
Hi @OmolemoBlessingLethuloe . What is the purpose of this change? These files are not used by the application anymore

OmolemoBlessingLethuloe:
The starter site: https://starter-adminportal-test.shesha.dev, is currently pointing to the localhost. Infrastructure notified that is the new implementation. Check here: https://dev.azure.com/boxfusion/Boxfusion%20DevOps%20Best%20Practices/_wiki/wikis/Boxfusion-DevOps-Best-Practices.wiki/969/Build-Pipeline-Fails-on-Rename-Config-Yaml-file-Rename-Item-Path-public-static-config-web.config.js#

IvanIlyichev:
@OmolemoBlessingLethuloe. These files are not used by the application anymore and this change doesn't make any sense. Currently application uses environment variables, you just need to set BACKEND_URL variable on the container/app service to point to the correct back-end

Issue #1114, opened 2024-02-27
Maintain the styling of the selected record even after clicking outside the table


hlayisanimazuze:
If you select a record on the table and click outside the table, the styling gets removed. The styling should persist even after clicking outside the table unless a user re-click the record to deselect/or click another record

**Steps to reproduce the behavior:**
1. Go to 'https://function-adminportal-test.shesha.dev/dynamic/Shesha/fileTest-table'
2. Click on a record within the table
3. Click anywhere outside the table
4. Actual Results: The color applied after selecting a record it gets removed

**Expected behavior**
1. If a user did not deselect the selected record or click another record, if they click outside of the table, the styling color should not get removed

PR #1135, opened 2024-03-05
persist selected row

nthangeniphumudzo:
Hi @IvanIlyichev . I was tasked to do this [item](https://github.com/shesha-io/shesha-framework/issues/1114). My changes are merely just making sure that we move away from the entire 'document' and just focus on the "table body" to avoid deSelect when clicking outside as per item stipulated.

### DIFF
--- a/shesha-reactjs/src/components/reactTable/tableRow.tsx
+++ b/shesha-reactjs/src/components/reactTable/tableRow.tsx
@@ -92,7 +92,9 @@ export const TableRow: FC<ISortableRowProps> = (props) => {
         setSelected(-1);
       }
     };
-    document.addEventListener('click', onClickOutside);
+    const tableRowsSection=document.getElementsByClassName(styles.tbody)[0] as HTMLElement;
+    
+    tableRowsSection.addEventListener('click', onClickOutside);
   }, []);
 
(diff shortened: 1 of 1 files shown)


IvanIlyichev (commented):
Hi @nthangeniphumudzo .This change is incorrect and doesn't fix the issue you mention. 
I think the initial logic was to clear the selection when a user clicks outside the table, but in some cases it makes sense to keep the selection.
The change you proposed just moves the event handler and to a first table in the DOM. 
If I understand correct this handler should be removed entirely, but if it's still required it should be attached to a correct DOM element, not to a first table, it will work incorrect if your page has multiple tables.

nthangeniphumudzo:
Thanks @IvanIlyichev for the insightful feedback. I will revise my implementation and communicate when I am done.

nthangeniphumudzo:
Hi @IvanIlyichev . I have made use of useRef distinguish each table and use it to demacate the inside/outside of each table. Please kindly review and feedback . Thanks

IvanIlyichev (commented):
Hi @nthangeniphumudzo. Row selection is handled wrong, I'll review it myself a bit later

Issue #560, opened 2023-11-20
If you set hidden property to 'True' on dataTableContext, all components within still shows


hlayisanimazuze:
When `datatableContext` hidden property is set to true and you render the form, the components within the `dataTableContext` still shows

**Steps to reproduce the behavior:**
1. From config URL: https://shesha-functionaltests-mainline-adminportal-test.azurewebsites.net/shesha/forms-designer/?id=2be1e350-078f-4d5c-aa76-5906a69af0bc, click on `datatableContext` component
2. On the properties panel, check the 'Hidden' property
3. Render the form

**Expected behavior**
The `dataTableContext` should respond accordingly to the property setting. If `hidden `property is `true`, therefore the component should be hidden

PR #1195, opened 2024-03-15
Hidding dataTableContext components when hidden is true 

czwe-01:
Addressed this [issue](https://github.com/shesha-io/shesha-framework/issues/560) with regards to dataTableContext showing components even when hidden is set to true


nthangeniphumudzo:
@czwe-01 can you just return null when the props is hidden to avoid the tenary chaining.

AlexStepantsov (approved):

PR #1212, opened 2024-03-24
added FluentValidationsOnEntityAsync method in DomainServiceExtensions

MaseruKganyago:


IvanIlyichev (changes requested):
Hi @MaseruKganyago. Please replace usage of Abp.Dependency.IocManager.Instance with StaticContext.IocManager

IvanIlyichev (approved):

Issue #537, opened 2023-11-20
The size of the 'Date Field' component reduces when styling is applied


hlayisanimazuze:
When styling is applied to the date field component, it unexpectedly reduces in size, leading to inconsistencies with other components

**Steps to reproduce the behavior:**
1. From this link:  https://function-adminportal-test.shesha.dev/shesha/forms-designer?id=fae89470-461b-4866-8e99-d3100a439f78, click on membershipStartDate component
2. On the properties panel, scroll down to style section and click on 'edit in code editor'
3. Return your CSS and save the changes
4. Preview the form(top right settings icon and eye icon)

**Expected behavior**
The component size should not change, unless the styling applied is to reduce its size

![date field](https://github.com/shesha-io/shesha-framework/assets/151041759/adfa72b5-8360-4834-8850-52618f1420d1)

PR #1219, opened 2024-03-25
fixed styled datePicker width

czwe-01:
Fixed [issue](https://github.com/shesha-io/shesha-framework/issues/537) of reduced width on datePicker when styling is added


nthangeniphumudzo:
@czwe-01  what happens when the user has custom style with width?

czwe-01:
> @czwe-01 what happens when the user has custom style with width?

There would be unwanted behavior with the width. 
Updated, thanks.

nthangeniphumudzo (approved):

Issue #1159, opened 2024-03-11
Provide additional options to configure the Panels component look & feel


ihouvet:
**Is your feature request related to a problem? Please describe.**
To make the panel component more suitable for use on public portals, additional configuration options should be provided.

**Acceptance Criteria**
1. The following properties should be configurable to provide more.
   1. Header background colour
   2. Body background colour
   3. Show/Hide Border styling
   4. Separator show/hide/styling (the bar/line between the header and body)
   5. Show/hide top bar
   6. Corner radius

7. The set of CSS classes that can be overidden to change the default look and feel should be clearly documented

Tshepiso-R:
NEed to create new issue for Corner Radius support

hlayisanimazuze:
Hi @Lukeybooi, can you please assist with the documentation for item number 2?

hlayisanimazuze:
Hi @Lukeybooi, this item is not complete

Requirement 2: The set of CSS classes that can be overidden to change the default look and feel should be clearly documented, mention you should document the CSS classes you can use to override the default look and feel. Please check this and provide the information. Suggestion you can also add info to the property tool tip/link to the documentation

Tshepiso-R:
Hey @Lukeybooi, please see @hlayisanimazuze's comment. Please provide documentation of the set of CSS classes that can be overridden to change the default look and feel
https://github.com/shesha-io/shesha-framework/issues/1159#issuecomment-2079036475

hlayisanimazuze:
https://dev.azure.com/boxfusion/Shesha%20Web%20v3.0/_wiki/wikis/Shesha-Web-v3.0.wiki/1107/Public-Portal-Styling

PR #1221, opened 2024-03-26
feat(collapsiblePanel): added additional properties

Lukeybooi:
- Simple Design property added
- Header Background Color property added
- Body Background Color property added
- Radius property added
- [Link](https://github.com/shesha-io/shesha-framework/issues/1159)

### DIFF
--- a/shesha-functional-tests/publicportal/src/components/layouts/mainLayout/index.tsx
+++ b/shesha-functional-tests/publicportal/src/components/layouts/mainLayout/index.tsx
@@ -3,14 +3,15 @@ import {
   ConfigurableApplicationComponent,
   CustomErrorBoundary,
   ISidebarMenuItem,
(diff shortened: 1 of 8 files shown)


IvanIlyichev (changes requested):
Hi @Lukeybooi. Please review the following:
1. Remove usage of static `.ant` prefix in components, it should be calculated using antd tokens
2. Review and clarify the requirements, I think some of these properties shouldn't be defined on the component level, it should be a part of theme instead
PS: antd 5 allows to customize the theme, token documentation available for each component ant the bottom of the corresponding component page, see example here: https://ant.design/components/collapse#design-token

IvanIlyichev:
PS: @Lukeybooi, please avoid usage of styled-components if the same functionality can be implemented using antd-style

Issue #539, opened 2023-11-20
The `style` property on `Dropdown` doesn't apply CSS with `value` dataSource


hlayisanimazuze:
When you apply styling on a dropdown component while `dataSource `type is value, the component does not respond to the styling

**Steps to reproduce the behavior:**
1. From the designer mode: https://function-adminportal-test.shesha.dev/shesha/forms-designer?id=65127122-2d1c-4a00-a0e1-17476e22dacb , click on the dropdown component
2. On the properties panel, scroll down and click 'dataSource' property
3. Select `values` option
4. Scroll down to `style `property and confirm the CSS script 
6. Preview the form
7. Actual Results: The styling does not apply on the component, however when the `dataSource` type is `referenceList`, the styling does apply

**Expected behavior**
The styling should apply according to the CSS properties added on the component

Similar and should be done in conjunction with #545

PR #1222, opened 2024-03-26
fix: styling dropdown with any data source 

czwe-01:
Addressed [issues](https://github.com/shesha-io/shesha-framework/issues/539) related to component not responding to the styling, and [issue](https://github.com/shesha-io/shesha-framework/issues/537) relating to size of datepicker changing when styling is applied.


nthangeniphumudzo:
@czwe-01  what happens when the user has a custom style that has width in it?

Issue #556, opened 2023-11-20
 Space size property is not functional on the 'Button Group' Component


Tshepiso-R:
The spacing between the buttons does not change when Space size is  changed

**Steps to reproduce the behavior:**

Form Link: https://shesha-functionaltests-temp-adminportal-test.azurewebsites.net/dynamic/shesha/test-js-funct?id=6db7f1ab-9155-477d-a9a5-04191365836c
1. Drag and drop the button group property
2. Configure at least two buttons
3. On the button group properties, change the Space size from middle to small, and from small to large.
4. Render the form

**Expected behavior**
When you change the space size the component should be adjusted according to the selected size.

PR #1237, opened 2024-03-28
space size issue/556

czwe-01:
Addressed the [issue](https://github.com/shesha-io/shesha-framework/issues/556) relating to button group not responding to space size.


nthangeniphumudzo (approved):
Mapped spaceSize to size

Issue #666, opened 2023-11-30
Button count label persists as if ButtonGroup buttons do not exist after adding ButtonGroup


hlayisanimazuze:
When a button is added to a `ButtonGroup (e.g., Button 1)`, and then another button `(Button 2)` is added, the button counter fails to increment correctly. Instead of displaying the number 2 for Button 2, it retains the number of the button within the `ButtonGroup (e.g., number 1)`

**Steps to reproduce the behavior:**
1. Go to 'https://function-adminportal-test.shesha.dev/shesha/forms-designer?id=ab1f059d-2476-41b9-ad13-0d1e1497ea85'
2. Click on `buttonGroup` component
3. On the properties panel, click on `Customise Button Group`
4. Click on` Add Group` button
5. Click on `Add New Item`
6. Click on `Add New Item `again
7. Actual Results: The second button retain the button number for the button inside the `buttonGroup`

**Expected behavior**
The button number should increment every time  

![image](https://github.com/shesha-io/shesha-framework/assets/151041759/b8870d11-b3bf-41e0-9371-69a954a4ab8c)

PR #1253, opened 2024-04-03
Button count label persists

czwe-01:
Fixed [issue](https://github.com/shesha-io/shesha-framework/issues/666) with persisting button label count


shesha-reactjs/src/providers/buttonGroupConfigurator/reducer.ts
```diff
@@ -28,8 +28,14 @@ const buttonGroupReducer = handleActions<IButtonGroupConfiguratorStateContext, a
       const newItems = [...state.items];
       const parent = state.selectedItemId ? (getItemById(newItems, state.selectedItemId) as IButtonGroup) : null;
       if (parent && parent.itemType === 'group') {
+        buttonProps.name = `button${parent.count + 1}`,
+        buttonProps.label = `Button ${parent.count + 1}`
         parent.childItems = [...parent.childItems, buttonProps];
-      } else newItems.push(buttonProps);
+        parent.count = parent.count + 1;
+      } else {
+        state.buttonCount  = state.buttonCount + 1;
```
AlexStepantsov: You cannot change the State directly
You can only return a new value of State from the method

AlexStepantsov (approved):

PR #1280, opened 2024-04-06
add german locals & fix indent

hakimdotdev:
- added german localization source file
- fixed indent


hakimdotdev:
I need the german locals to be able to use shesha

Issue #627, opened 2023-11-23
Embed the 'Entity Reference' component to 'Autocomplete' component


hlayisanimazuze:
1. Replace the available `Quickview` property with `Entity Reference`
2. The Panel name should be `Entity Reference`
3. The checkbox should be renamed to `Use Entity Reference`
4. The checkbox should be defaulted to false
5. When `Use Entity Reference` is set to true, entity referencing properties should display allowing a user to configure accordingly. This should support all the options we have on the Entity Reference component
6. This entity reference functionality is for read only mode
7. When navigating to a edit mode, autocomplete component should be displayed

![image](https://github.com/shesha-io/shesha-framework/assets/151041759/c5fe81e9-572e-4575-9d50-62c158df9d7e)

![image](https://github.com/shesha-io/shesha-framework/assets/151041759/1d3140af-3ccf-43ec-b842-06af07df800a)

PR #1282, opened 2024-04-08
James/en/627

James-Baloyi:
Embed Entity Reference component to Autocomplete component

https://github.com/shesha-io/shesha-framework/issues/627


AlexStepantsov:
please, add a component migration from old to new properties

James-Baloyi:
@AlexStepantsov can you elaborate more on that? Like which properties do I need to migrate?

AlexStepantsov:
@James-Baloyi 
Old QuickView properties to the new EntityReference properties.

The migration should preserve the behavior of components that are already used on forms, as it would have been before the changes
For example, set the default of `entityReferenceType` to "QuickView" mode. Copy `quickviewFormPath`, `quickviewDisplayPropertyName`, etc to the new `formIdentifier`, `displayProperty`,...

btw, why didn't you remove the old properties (`quickviewFormPath`, `quickviewDisplayPropertyName` and other)?

James-Baloyi:
@AlexStepantsov Thanks. I've added Quickview properties to migrations and removed unused ones from the Autocomplete props object

AlexStepantsov:
@James-Baloyi 
You should not change existing migration, you have to add a new migration.

I will made changes

AlexStepantsov:
Hi @James-Baloyi 

PLease check follow:
1. Your changes don't work at all.
2. You added settings (markup), but there are no corresponding changes in interfaces and models
3. You did not implement main requirement - Embed the 'Entity Reference' component. Mainly you just changed settings and add 'EntityReference' component to the markup but it doesn't work (including the added settings are not used)
4. You left the old settings in interfaces and models and did not change their use in the component code (accordingly, the old settings affect the behavior of component)
5. And lastly, there is no migration from old settings model to the new one... (you added something, but it seems like you did it at random)

Issue #583, opened 2023-11-21
The formatting applied on nested entities configurations do not apply on subform


Tshepiso-R:
The applied formatting on nested entity configurations is not being reflected correctly within a subform. The price has formatting applied on the Book Entity, however, when pulling the data it does not come with the formatting

**Steps to reproduce the behavior:**

Form URL: https://function-adminportal-test.shesha.dev/dynamic/Boxfusion.SheshaFunctionalTests.Common/subject-details?id=cc9ff707-8cb3-4976-849e-0034e34dff47
1. Click the above link
2. Check the data pulled from a sub-form

**Expected behavior**
Formatting applied to nested entity configurations should be correctly displayed within the subform.

![image](https://github.com/shesha-io/shesha-framework/assets/151129946/64d2ea91-1ce7-452e-a012-f9e914fdead4)

Tshepiso-R:
Lukeman Elabor
[commented Oct 2](https://dev.azure.com/boxfusion/Shesha%20Web%20v3.0/_backlogs/backlog/Shesha%20Web%20v3.0%20Team/Stories/?workitem=49261#4261906)
 
Hi @hlayisanimazuze  please note that the metadata is only applicable on the table. No other entity requires the types metadata. Please set this as a new feature not a bug if that is the case

PR #1345, opened 2024-04-17
Thulasizwe/fix/583

czwe-01:
Addressed [issue](https://github.com/shesha-io/shesha-framework/issues/583) that caused the formatting applied on nested entities configurations to not apply on subform


nthangeniphumudzo:
@AlexStepantsov  please kindly have a look at my changes . The Subform Fields were not receive the appropriate "format" .  So the main change I did are:
1. the creation of a hook "useEntityProperties" that prioritise the subForm modelType and if not available it the resort to the main formSettings to get properties. My ultimate goal was to get the right properties everytime.
2.I have updated the "getDataProperty" to break down the propertyName to the exact name from the "entity".
3.I have assign the globalFormat to NumbrerField, timeField, DateField which can be overrode by the component level "format".

Issue #1318, opened 2024-04-12
Add `name` property on formConfiguration update endpoint


hlayisanimazuze:
When updating the form `name`, the information does not get saved because the `propertyName `field is not available on the update endpoint.

**Steps to reproduce the behavior:**
1. Go to 'https://function-adminportal-test.shesha.dev/dynamic/shesha/form-details?id=a6db7773-93c9-43e4-b33b-00f171ccd032'
2. Click on 'edit'
3. Update 'name' field
4. Click on 'Save' , refresh the form and check the value displaying on the name field
5. Actual Results:  The information was not updated

Backend
There are the properties available on the formConfiguration/Update ndpoint
![image](https://github.com/shesha-io/shesha-framework/assets/151041759/19891910-00d7-40ce-8738-3e0cf2af6fc6)

Endpoint executed on the frontend
<img width="952" alt="image" src="https://github.com/shesha-io/shesha-framework/assets/151041759/7afcebca-ea0d-4a4b-89ef-a29e38eb76d0">

**Expected behavior**
The name property should be added to give users flexibility to update the form name

PR #1405, opened 2024-04-25
Adding name property on formConfiguration update endpoint

OmolemoBlessingLethuloe:
https://github.com/shesha-io/shesha-framework/issues/1318


IvanIlyichev (changes requested):
Hi @OmolemoBlessingLethuloe. Validation check is incorrect, please recheck. Current row shouldn't be compared with itself

Issue #1468, opened 2024-05-08
The payload sends the exact format received from the component which differs to what the backend is expecting


hlayisanimazuze:
If a user configure the dateTime format different to what the backend is expecting, the payload sends the exact format received from the component and the backend rejects it. 

**Steps to reproduce the behavior:**
1. Go to 'https://function-adminportal-test.shesha.dev/dynamic/TestModule/fileTest-table'
2. Open developer console
3. Click on 'Default' button
4. Enter the details and submit
5. Check the payload
6. Actual Results: The payload sends the exact format received from the component

**Expected behavior**
Component should display the value in the configured format but store it as a date and send it to the back-end always in the ISO format

![image](https://github.com/shesha-io/shesha-framework/assets/151041759/64f795a6-6dc2-4c4e-b0f4-62f3d18895f1)

![image](https://github.com/shesha-io/shesha-framework/assets/151041759/cf91e553-3a02-49c9-83ff-86452cc3a910)

This also impact the dynamic filtering: https://function-adminportal-test.shesha.dev/dynamic/Boxfusion.SheshaFunctionalTests.Common/testform1-create?mode=edit. Select payment date and check the endpoint

PR #1479, opened 2024-05-09
Phumudzo/fix/1468

nthangeniphumudzo:
Hi @IvanIlyichev . Please kindly assist with this PR addressing[ this issue](https://github.com/shesha-io/shesha-framework/issues/1468). 

My Thinking process here was that the value passed to the DatePicker should always be in ISO 8601 and allow it to be formatted by the pickerFormat.

Please review and advise on this approach.


shesha-reactjs/src/designer-components/dateField/datePickerWrapper.tsx
nthangeniphumudzo, line 1: removed localValue since it returning the same date as datesString which is simple to manipulate and format
  I thought of retaining it for user manipulation but it eventualklly returns the same date as dateString

IvanIlyichev:
Required changed have been implemented by @AlexStepantsov as part of another PR

Issue #1483, opened 2024-05-09
Import/Export forms loader continues loading non stop even after the endpoint has successfully completed


hlayisanimazuze:
- After successfully completing the import endpoint, the loader indicator persists in loading even though the packages have been imported onto the system
- If the action get cancelled such as form export/import/publish/setready/cancel version, it does not stop loading

**Steps to reproduce the behavior:**
1. Go to 'https://function-adminportal-test.shesha.dev/dynamic/shesha/forms'
2. Open your developer console
3. Click on 'import' button
4. Select the json package. Use the attached file
5. Click on import button
6. Actual Results: After few seconds, a toast message gets displayed and the endpoint completes successfully, while the loader still loads

Cancel the action
1. Click on export/import
2. After the dialog shows, click on cancel or 'x' to close the dialog
3. Actual Results: The loader does not stop loading

NB: Can this be a global fix, to also fix any item that behaves that way

**Expected behavior**
After the endpoint successfully completes or the action gets interrupted, the loader indicator should stop loading 

![image](https://github.com/shesha-io/shesha-framework/assets/151041759/21ea0960-3e8f-4172-9d85-edddcbfff343)

https://github.com/shesha-io/shesha-framework/assets/151041759/d96d0452-9c46-4ec5-a266-79c6878e130f

PR #1507, opened 2024-05-13
Phumudzo/fix/1483

nthangeniphumudzo:
Hi @IvanIlyichev . I am addressing [issue](https://github.com/shesha-io/shesha-framework/issues/1483) raised here. This actions were previous overlooked.


IvanIlyichev (changes requested):
Hi @nthangeniphumudzo. If the configurable action fails we should reject the promise with an error message, not resolve it with false result.

Issue #1166, opened 2024-03-12
Datefield formatting (DD-MM-YYYY) data doesn't match endpoint data/what has been selected


hlayisanimazuze:
If this DD-MM-YYYY formatting is used on the table  `dateField ` display component, details view `dateField `component and entity configuration(details only), it display data that does not match what the endpoint has. This issue occurs when
 using the following:

- Inline adding and display `dateField `component
- Details view updating the formatting property on the `dateField` and create
- When the formatting is applied on the `entity configuration`, the details view still displays the wrong data than what's received from the endpoint

**Steps to reproduce the behavior adding:**
1. Go to 'https://function-adminportal-test.shesha.dev/dynamic/TestModule/testTable-member'
2. Click on 'membership Start Date' and select a date
3. Actual Results: The displayed date does not match what is being selected

**Steps to reproduce the behavior displaying:**
1. Go to 'https://function-adminportal-test.shesha.dev/dynamic/TestModule/testTable-member'
2. Search for conversation booster
3. Check membershipState date column and compare with that data from the endpoint
4. Actual Results: Both values do not match

**Steps to reproduce the behavior details view:** (This also applies when the formatting is applied on entity configuration)
1. Go to 'https://function-adminportal-test.shesha.dev/dynamic/TestModule/testBlank-member?id=4d5e9904-f65b-402f-982e-7b7370854a9c'
2. Check membershipState date column and compare with that data from the endpoint/Edit the component and select a new value
4. Actual Results: Both values do not match

**Expected behavior**
1. When the formatting is applied on all the above mentioned use cases, the formatting should not change the data but display what is received from the endpoint

<img width="946" alt="image" src="https://github.com/shesha-io/shesha-framework/assets/151041759/ae0acf8f-4fb2-4f6b-b67a-0d7a6ccec00f">

PR #1527, opened 2024-05-15
Phumudzo/fix/date format

nthangeniphumudzo:
Hi @IvanIlyichev . I revisted the issue and here is my new approach.

I know you don't want manipulation of the "localValue" before transimission/submission, but in this instance I am not change it I am just making sure the data sent is aligned with UTC and avoiding that transformation happening in .Json() on submission where we won't have control all together.

here is the link to the [issue](https://github.com/shesha-io/shesha-framework/issues/1166)


IvanIlyichev (changes requested):
Hi @nthangeniphumudzo. Range pickers are broken. Also we should use start of the periods (e.g. start of the day for date only, start of the week for weeks etc)

Issue #1326, opened 2024-04-14
Navigation ‘links’ and buttons should behave as regular URL based links to take advantage of native browser functionality


ihouvet:
### Overview
Currently navigation ‘links’ and buttons in Shesha do not behave as normal links rather they execute JavaScript, which causes the navigation. The problem with this is that we are unable to take advantage of native browser functionality, such as mouse, clicking, and opening in a separate tab.
Some users, including myself, find this extremely useful, but it’s not possible in Shesha.
This is in particularly limiting for the main menu items and the drill down/action columns on table views typically used to view or edit details of an entity.

### Acceptance Criteria
1. Hovering over a main menu navigation item should display the URL. The item will navigate the user to in the browser footer as per standard behavior for a link.
   1. Right mouse, clicking on a main menu navigation item should offer the standard options to open the destination in a new tab or a new window
1. Hovering over a navigation column (e.g. view details or edit column) on a data table should display the UR in the browser footer as per standard behavior for a link.
   1. Right mouse, clicking on a the navigation column should offer the standard options to open the destination in a new tab or a new window

### Current Behavior
<img width="1366" alt="image" src="https://github.com/shesha-io/shesha-framework/assets/21078025/180e8541-a170-406f-ab6c-cf719bc301e9">

### Required Behavior
<img width="1366" alt="image" src="https://github.com/shesha-io/shesha-framework/assets/21078025/7f3100df-6d65-4923-bb3d-0b0ea8f4feae">

hlayisanimazuze:
Hi @James-Baloyi , please set this card to `readyForTesting `once all the items are resolved

hlayisanimazuze:
Closing this item. Any issues arising we will be linked to this card. What's outstanding is #1598

PR #1672, opened 2024-06-10
fixed button crashing when query parameters are not configured

James-Baloyi:
Relating to #1326


IvanIlyichev (approved):
Hi @James-Baloyi. I approve your change only as a quickfix, we'll discuss your code a bit later

Issue #710, opened 2023-12-05
Allow Configurator to specify Required permissions at a view level


hlayisanimazuze:
1. The Configurator should be able to specify 'Required permissions' at the view level
2. If a user attempts to access the view without the required permissions a 'Unauthorised' message should be displayed

hlayisanimazuze:
The implementation has been completed; however, there are limitations in the current implementation. Specifically, subforms rendered or dialogs within the details views are not behaving as expected. A separate card has been created to address this limitation once the review of the form lifecycle has been completed #1737

PR #1736, opened 2024-06-20
User/omolemo/view permissions

OmolemoBlessingLethuloe:
Related issue: https://github.com/shesha-io/shesha-framework/issues/710


IvanIlyichev (commented):
note: the behaviour should be changed after review of the form/page lifecycle. When a form is rendered as a main content of the page it should show a warning (as implemented in current PR), if the form is rendered as part of page and user has no required rights it should be hidden

PR #1858, opened 2024-07-17
Public Configuration

Lukeybooi:
- Created login Async
- Added Login Migration
- Added Public Portal Migrations


IvanIlyichev (changes requested):
Hi @Lukeybooi. 
1. navigation shouldn't be done using window object, we should use router. Use **Navigate** action instead of **Execute Script**, see attached image
image

2. IAuthActionsContext.loginUser - I didn't find any usages, loginUserAsync is used instead. Please remove duplicate
3. Login button should be pressed automatically by press enter on the password field. - ignore for now, I think we need a separate car for it because it requires of events or submit actions

**NOTE**: Please disable prettier on your dev environment, it makes PRs review much more complex. We'll enforce auto-formatting or checking on the repo level later

IvanIlyichev (approved):

Issue #1378, opened 2024-04-23
Extend the theme page to support global styling theme


hlayisanimazuze:
- Add a Styling section on the theme page
- On the section, add the outstanding approved AntD styling appearance such as corner radius
- Use this link for your reference:  https://ant.design/theme-editor#Style

PR #1859, opened 2024-07-17
added border radius to theming

James-Baloyi:
#1378


IvanIlyichev (changes requested):
Hi @James-Baloyi. Please set default value for radius to 8px to keep existing styling.

Issue #1878, opened 2024-07-19
 Enhance input components with additional styling properties


Tshepiso-R:
**Description:**
We need to enhance the Input components by adding the following properties to improve its customization and styling options. The relevant components have been listed below:

**Properties to Add:**

1. Width
2. Height
3. Styling:
4. Background Color
5. Font Size
6. Font Color
7. Font Weight
8. Border (width, color, solid, dotted, dashed)
9. Border Radius
10. Icon Position
11. Additional Types:

    - Text
    - Number
    - Email
    - Password

12. StylingBox:
    - Margin
    - Padding
    
**Duplicates For:**
1. Autocomplete
2. Dropdown
3. Date Picker
4. Time Picker
5. Number Field
6. Textarea
7. Autocomplete Tags (adjust input height)

hlayisanimazuze:
Hi @James-Baloyi, The work item lists multiple components that should be updated to have the same properties. From the list, only Textfield, Autocomplete, and Text Area have been implemented.

If there is a reason why not all components should be updated, please add a comment so we can update the item accordingly. Currently, we can't close the item since the requirements are not fully implemented.

ihouvet:
Need to verify this has been catered for in the properties panel rework

zukidlomo:
Closing issue because the requested additions have been implemented.

PR #1926, opened 2024-07-30
James/e/1878autocomplete

James-Baloyi:
#1878


IvanIlyichev (changes requested):
Hi @James-Baloyi. Border style should work taking in account the **Hide border** and it should be implemented by the same way for all components affected by your changes. The **Hide border** should be above other border settings which tweak border style and these settings should be hidden when the **hide border** is switched on. Please review other components too

James-Baloyi:
@IvanIlyichev the requested changes have been committed

IvanIlyichev (changes requested):
Hi @James-Baloyi. 
1. Border radius is still visible when the Hide border is switched on. 
2. Please revert back **variant** property, all components should behave by the same way

James-Baloyi:
@IvanIlyichev the reason border radius is still visible is the same as the one here https://github.com/shesha-io/shesha-framework/pull/1927

I've added the default border radius and reverted variant to the previous implementation.

PR #1927, opened 2024-07-30
James/e/1878textfield

James-Baloyi:
#1878


IvanIlyichev (changes requested):
Hi @James-Baloyi. See comments here https://github.com/shesha-io/shesha-framework/pull/1926

IvanIlyichev (changes requested):
Hi @James-Baloyi. **Border radius** is still visible when the **Hide border** is switched on

James-Baloyi:
@IvanIlyichev I kept it on because it's still applicable even when the border is not shown. For example you could want to create an input field like this one 
image

Issue #1530, opened 2024-05-16
The configuration environment (admin portal) should allow configurators to configure various client specifics


hlayisanimazuze:
The App Settings framework and page needs to be enhanced to allow managing Front-end specific settings in addition to general settings currently supported.

Expand the Settings Page:
- Include a dropdown next to the title (see illustration below) that allows the user to select 'General' to view general settings (current behaviour) and lists all registered front-end applications to allow the user to manage front-end specific settings.

General Settings:
- By default, the settings page should display all the General settings that are supported across all frontends.

Frontend-Specific settings:
- The user should be able to select any of the registered front-ends for the application
- Upon selecting a frontend application from the dropdown, the front-end specific settings should be listed
- By default the following options should be displayed to allow a configurator to configure settings for the selected front-end:
  - Default url: this is the url the user should be redirected to if the user is not authenticated and does not specific a specific page i.e. just enters the domain e.g. https://www.boxfusion.co.za  . Currently the user is directed to the login page but we need the ability to configure this.
  - Theme - Move the theme configuration page here
  - Main Menu 

![image](https://github.com/shesha-io/shesha-framework/assets/21078025/c4c1205d-efc7-491e-85c6-aa5966100b2e)

PR #1950, opened 2024-08-02
APIs should exist to store user specific settings and preferences V2

OmolemoBlessingLethuloe:


AlexStepantsov (approved):
Approved and will be updated by Alex Stephens for 'The configuration environment (admin portal) should allow configurators to configure various client specifics #1530'

Issue #1957, opened 2024-08-05
Unable to revert to value property after clicking JS in Key Information Bar component


miracleMa:
**Describe the bug**
when you click on the JS  in any property on the key information bar component, it will not allow you to revert to the value property configurations. (in all the properties except the Hidden property)

**To Reproduce**

1. Click"https://function-adminportal-test.shesha.dev/shesha/forms-designer?id=d17ac689-d027-4ff1-a442-04e8cae65102"
2. Drag and drop the key information bar component,
3. Navigate to Properties 
4. Click  on the "JS "property 
5. Click on the "value" property

**Actual results**
It won't  allow you to revert to the value property configurations
**Expected behavior**
it should change back to the value property configurations.
**ScreenRecord**
[function-adminportal-test-shesha-dev-shesha-forms-designer-id-7fab457e-ebec-49df-9948-538ee07f3cca.webm](https://github.com/user-attachments/assets/fb98a727-303f-4d36-b9bc-b1002caa7240)

PR #1979, opened 2024-08-12
Clicking on js button on property configurations toggles between the modes

czwe-01:
This PR addresses the [issue](https://github.com/shesha-io/shesha-framework/issues/1957) when you click on the JS in any property on the key information bar component, it will not allow you to revert to the value property configurations. (in all the properties except the Hidden property).


AlexStepantsov:
incorrect implementation

PR #1986, opened 2024-08-13
fix: Check if the permission is not already created 

teboho:
When we export configured permission(s) and then try to import them, either including them when we export forms or just exporting the configured permissions alone, the backend tries to do the operation twice which ends up breaking the first build. 

This change will make sure that if a valid permission of that name already exist, it does not get created twice.


teboho:
Resolves https://github.com/shesha-io/shesha-framework/pull/1986

AlexStepantsov:
the original problem is not fixed

PR #1992, opened 2024-08-13
Removed the use of LoginWrapper as it does not exist anymore

teboho:
Use of LoginWrapper in login page breaks the pipeline because it causes the build for the starter to build


teboho:
Got fixed

Issue #1847, opened 2024-07-13
The sizable columns component crashes when adding a new column


hlayisanimazuze:
The component initially comes with two columns pre-configured. However, when an additional column is added, the component crashes

**Steps to reproduce the behavior:**
1. Go to any designer form and drag and drop the sizable columns component from the builder widgets
2. On the properties panel, click on 'configure columns' button
3. Click on 'Add column' button
4. Click 'Ok'
5. Actual Results: Cannot read properties of undefined(reading 'a')
6. Click on 'Try again'
7. Actual Results: The error disappears

**Expected behavior**
Upon clicking ok, the column should be added without crashing

![image](https://github.com/user-attachments/assets/d9f8ebbd-12a5-4b18-a1fc-939916967c0b)

PR #1997, opened 2024-08-14
Fix/le/sizable columns

Lukeybooi:
- The sizable columns component crashes when adding a new column #1847


IvanIlyichev (changes requested):
Hi @Lukeybooi.
1. Your changes break usage of resizeable panels in the form designer itself
2. Timeouts (e.g. usage of setTimeout) only hide errors which should be properly fixed
3. SizableColumnsList components must not change value on the fly, it should call onChange only when the user clicks `Ok` on the modal. Review this component first, it may solve the issue
4. SizableColumnsList is a copy of legacy editor, it should be replaced with generic ListEditor (see example [here](https://github.com/shesha-io/shesha-framework/blob/3bc6ab0d8d79815423a5af24a39cf7626bf967ad/shesha-reactjs/src/components/dataTable/sortingConfigurator/index.tsx#L21))

Issue #1999, opened 2024-08-14
Add shadow control to container component


Tshepiso-R:
**Feature Description:**

1. Introduce a shadow effect option for the container component.
2. The shadow will have predefined styling 
3.Users can enable or disable the shadow effect through a toggle switch in the properties panel.

**Properties Panel:**

1. Add a toggle control labeled "Shadow" to the properties panel of the container component.
2. When enabled, the shadow will be applied to the container.
3. When disabled, the shadow will be removed.

hlayisanimazuze:
PR

hlayisanimazuze:
Further enhancements to the shadow style property will be addressed by the changes pushed to the properties panell: https://github.com/shesha-io/shesha-framework/issues/2264

PR #2008, opened 2024-08-15
simple shadow implementation on container

James-Baloyi:
#1999


shesha-reactjs/src/designer-components/container/interfaces.ts
```diff
@@ -164,6 +164,8 @@ export interface ICommonContainerProps {
   borderStyle?: string;
   borderRadius?: string | number;
   overflow?: string;
+  allowShadow?: boolean;
+  shadowStyle?: string;
```
AlexStepantsov: Please use one dropdown shadowStyle - None, Above, Below
And don't forget to add a migration to set shadowStyle = none

James-Baloyi: Will do

AlexStepantsov (approved):

Issue #1998, opened 2024-08-14
Inconsistent font sizes in configurable header profile dropdown


Tshepiso-R:
**Description:**
The font sizes within the configurable header, specifically in the profile dropdown, are inconsistent across various components. 

**Actual Result:**
The font sizes within the profile dropdown are inconsistent.

**Expected Result:**
All text elements within the profile dropdown should have consistent font sizes as per the design specifications.

hlayisanimazuze:
PR

James-Baloyi:
@Tshepiso-R this functionality is handled by the configurable header component, we can close the item.

PR #2011, opened 2024-08-15
added font size to profile dropdown component

James-Baloyi:
#1998


James-Baloyi:
The relevant component is now configurable and no longer relies on hardcoded styles

PR #2046, opened 2024-08-20
Updating table view template to reference the new table filter compon…

OmolemoBlessingLethuloe:
…ent and closing item model by default


IvanIlyichev (changes requested):
Hi @OmolemoBlessingLethuloe. The form you exported is not a template and it contain leading space in the name. Also please make sure that your template is based on the latest version of table-view template (see package20240815_1811.shaconfig, I updated it earlier)

PR #2048, opened 2024-08-21
Maseru/header config setting

MaseruKganyago:
allowing flexibility of header-config-form-id on project-level


IvanIlyichev (changes requested):
Hi @MaseruKganyago. Please move the **headerConfiguration** to the props of the **MainLayout**, it shouldn't be a part of the applicaton. Also rename it to **headerFormId**

shesha-reactjs/src/providers/sheshaApplication/index.tsx
```diff
@@ -69,6 +70,7 @@ export interface IShaApplicationProviderProps {
    */
   applicationKey?: string;
   getFormUrlFunc?: (formId: FormIdentifier) => string;
+  headerConfiguration?: IPersistedFormProps;
```
IvanIlyichev: Hi @MaseruKganyago. Please remove unused property

IvanIlyichev (approved):

Issue #2038, opened 2024-08-19
Recent changes in `Button` and `Button Group` impact layout and functionality


Tshepiso-R:
### Description:
Recent updates to the button and button group components have caused several issues affecting the layout and functionality of existing configurations.

### Issues and Fixes:

1. **Layout Changes:**
   - **Issue:** The button layout of existing button configurations has been altered.
   - **Image:**
     ![Layout Issue](https://github.com/user-attachments/assets/550eb22f-fdd1-4d25-95ab-be80dbe4f9c1)
   - **Expected Layout:** [Expected Layout](https://function-v0-37-adminportal-test.shesha.dev/)

2. **Configuration Reflection in Dialog:**
   - **Issue:** Changes to border type, border width, and background color are only visible after closing and reopening the dialog.
   - **Fix:** Implement real-time updates so changes are reflected immediately within the dialog.

3. **Clear Button for Background , Border and Color Properties on button group:**
   - **Issue:** Missing clear button for border color and color properties.
   - **Image:**
     ![Clear Button Missing](https://github.com/user-attachments/assets/2065cfb0-05b5-44eb-bb88-c5d9cee1c83a)
   - **Fix:** Add a clear button for these properties to allow users to reset their choices easily.

4. **Width Display in Dialog:**
   - **Issue:** The width displayed in the dialog does not reflect the default button widths and defaults to block.
   - **Image:**
     ![Width Display Issue](https://github.com/user-attachments/assets/37d79f1a-75dd-4a21-bb6b-5230cb0b149c)
   - **Fix:** Update the dialog to show the correct width. 
   
5. **Button Group Size Setting:**
   - **Issue:** The size setting is not functioning for button groups.
   - **Fix:** Ensure that the size setting applies correctly to all buttons within the group.

6. **Button Type in Groups:**
   - **Issue:** Button type configuration is not applying to buttons within a group.
   - **Configured Button:**
     ![Configured Button](https://github.com/user-attachments/assets/5fd4b365-bad2-4be5-a098-ef5a6b060b42)
   - **On Render:**
     ![On Render](https://github.com/user-attachments/assets/75386fd7-da07-4492-a090-6e5185d1fd7e)
   - **Fix:** Ensure that the button type configuration is properly applied to each button in the group context.

7. **Redundant Button Style Properties:**
   - **Issue:** Two properties for Button Style are redundant and perform the same function.
   - **Image:**
[...]

PR #2169, opened 2024-09-18
Thulasizwe/en/1877: button and button group styling

czwe-01:
Addressing [issue](https://github.com/shesha-io/shesha-framework/issues/2038). Recent updates to the button and button group components have caused several issues affecting the layout and functionality of existing configurations.

- [ ]  The button layout of existing button configurations has been altered
- [ ]  Changes to border type, border width, and background color reflect immediately
- [ ] Colorpicker has clear button
- [ ] Configurator reflects the correct default button width
- [ ] Button group size applies to all buttons within the group
- [ ] Button type applied to buttons in a group.
- [ ] Hidden property re-added

### DIFF
--- a/shesha-reactjs/src/components/buttonGroupConfigurator/buttonGroupItem.tsx
+++ b/shesha-reactjs/src/components/buttonGroupConfigurator/buttonGroupItem.tsx
@@ -34,7 +34,31 @@ export const ButtonGroupItem: FC<IButtonGroupItemProps> = ({ item, actualModelCo
   const { styles } = useStyles();
   const actualItem = useMemo(() => getActualModel({ ...item, actionConfiguration }, actualModelContext)
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    , [{ ...item }, actualModelContext]);
+    , [
+      item.icon,
+      item.label,
+      item.tooltip,
+      item.iconPosition,
+      item.size,
+      item.buttonType,
+      item.borderColor,
+      item.borderRadius,
+      item.height,
+      item.width,
+      item.backgroundColor,
+      item.fontSize,
+      item.fontWeight,
+      item.color,
+      item.borderStyle,
+      item.borderWidth,
+      item.readOnly,
+      item.style,
+      item.block,
+      item.danger,
+      actionConfiguration,
+      actualModelContext,
+      actualModelContext
+    ]);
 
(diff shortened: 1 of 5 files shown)


shesha-reactjs/src/components/buttonGroupConfigurator/buttonGroupItem.tsx
```diff
@@ -34,7 +34,7 @@ export const ButtonGroupItem: FC<IButtonGroupItemProps> = ({ item, actualModelCo
   const { styles } = useStyles();
   const actualItem = useMemo(() => getActualModel({ ...item, actionConfiguration }, actualModelContext)
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    , [item.label, item.icon, item.tooltip, item.name, actualModelContext]);
+    , [{ ...item }, actualModelContext]);
```
AlexStepantsov: You must not do that, your implementation just always creates a new object and the recalculation will be performed at each rendering regardless of changes in properties

shesha-reactjs/src/components/buttonGroupConfigurator/properties.tsx
AlexStepantsov, line 1: Why are these changes here?

shesha-reactjs/src/providers/form/models.ts
```diff
@@ -189,6 +190,8 @@ export interface IConfigurableFormComponent
   noDataSecondaryText?: string;
 
   permissions?: string[];
+
+  allowClear?: boolean;
```
AlexStepantsov: Why is this property here? It is clearly specific to some component (I assume ColorPicker). What exactly in ALL components should be allowed to be cleared?

shesha-reactjs/src/components/buttonGroupConfigurator/buttonGroupItem.tsx
czwe-01: When adding item as a dependency as it is, changes did not reflect the applied styles immediately since , till I started listing the style properties, which seemed a bit excessive because there were many properties to list, but now that you explain it, I see that my approach is not ideal and can affect performance.

shesha-reactjs/src/providers/form/models.ts
czwe-01: I noticed that it was supported in a couple of input components e.g. Input, datePicker, colorPicker, autocomplete, and Select, so I thought moving it up was a good idea.  Now I see it's not because very few components support it.

AlexStepantsov (approved):

PR #2220, opened 2024-10-02
Charts component

teboho:
# `Charts` component v0

> Configurable chart component

https://github.com/user-attachments/assets/f50a00e8-cca1-4044-a993-bead3c98be68  

## Pivot / Simple charts  

This is the early, usable version that shall get better with time passively.  
The code will get cleaner as well : | 

- Bar
- Pie
- Line


teboho:
https://github.com/shesha-io/shesha-framework/issues/2221

IvanIlyichev (changes requested):
Hi @teboho. Make the following changes:
1. Use MetadataDispatcher to get metadata
2. Use ReferenceListDispatcher for reference lists fetching
3. Use types as much as possible, I see a lot of usages of 'any' in the data fetching methods, it makes the code fragile

IvanIlyichev (approved):

PR #2279, opened 2024-10-10
Added API to publish all forms at once

mpendulontshangase:
image
image (1)


IvanIlyichev (changes requested):
Hi @mpendulontshangase. **ConfigurationItemAppService** is a common application service for all types of configurable items. The action that publish all forms shouldn't be there, it should be a part of **FormConfigurationAppService** and the name of action should be **PublishAllAsync** (without **forms** and with Async suffix)

IvanIlyichev (changes requested):
Hi @mpendulontshangase. Please update form configuration to use a correct API endpoint

IvanIlyichev (approved):

PR #2293, opened 2024-10-15
Charts component update

teboho:
- if a parameter is a reflist property, get the names/labels of the different reflist items so that you show those on the labels
- set buttons to primaryColor
- move entity/data settings to the top
- remove name and description settings


IvanIlyichev (changes requested):
Hi @teboho. Build failed because of unused imports in one of files. Make sure that `npm run build` works on your side

IvanIlyichev (approved):

Issue #2315, opened 2024-10-22
Datalist form info does not disappear after 3 seconds in edit mode


Tshepiso-R:
**Describe the bug**
When you are in edit mode and navigate to a form that contains a datalist, the form info for the datalist displays but does not disappear after 3 seconds. The main form's info disappears as expected, but the datalist's info only disappears after you hover over the datalist section, which should not happen.

**To Reproduce**
1. Go to 'https://functional-adminportal-test.azurewebsites.net/dynamic/TestModule/driverList-table'
2. Enable Edit Mode.
3. Refresh the form and keep the cursor on the browser’s refresh button.
4. Observe that all form info on the page disappears after 3 seconds except for the datalist form info, which persists until the datalist section is hovered over.

**Expected behavior**
- The datalist form info should disappear automatically after 3 seconds, along with the rest of the form info, without requiring the user to hover over the datalist section.

**Actual behavior**
- The datalist form info remains visible until the user hovers over the datalist section, while the main form info disappears after 3 seconds as expected.

**Screenshot:**
![image](https://github.com/user-attachments/assets/c40195de-4c69-4c18-b3dd-04075006f4a8)

Related Item: #1517

hlayisanimazuze:
Hi @James-Baloyi , the issue is still persisting on the datalist component. I've cleared my cache and still get the same results

- Navigate to: https://functional-adminportal-test.azurewebsites.net/dynamic/TestModule/newmode-details?id=1b82e19b-a799-4401-9971-00b7bcd7ba4b
- Switch to edit mode
- After 3 seconds, the form info will disappear
- Refresh the page and keep your mouse far away from the form/keep it on the refresh icon
- Observe all the forms info disappearing except the one rendered within the dataList. See attached recording:  https://jam.dev/c/de0071ea-d0de-455d-9fb4-8233c730947c

James-Baloyi:
@hlayisanimazuze the issue has been addressed and we have updated the behavior of the 3 second timer to only trigger when the user clicks on the 'Edit Mode' toggle. This means form info blocks only go away after an explicit user action instead of when the page is reloaded and there's no explicit action to trigger the behavior.

This is a best practice requested during the review process.

PR #2325, opened 2024-10-24
removed conditional check before hiding formInfo panel

James-Baloyi:
#2315

### DIFF
--- a/shesha-reactjs/src/components/appConfigurator/editModeToggler.tsx
+++ b/shesha-reactjs/src/components/appConfigurator/editModeToggler.tsx
@@ -7,14 +7,14 @@ import { useStyles } from './styles/styles';
 export interface IAppEditModeTogglerProps { }
 
 export const AppEditModeToggler: FC<IAppEditModeTogglerProps> = () => {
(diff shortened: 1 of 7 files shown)


IvanIlyichev (changes requested):
Hi @James-Baloyi. The code requires refactoring. Current implementation has multiple issues which make it unstable and complicate code readability.

1. If your code requires timeouts it should be set on the action (e.g. onClick set visible and start timeout to hide) but not as part of the useEffect.
2. avoid unneeded useEffects in general, current logic of the control is unpredictable especially taking in mind that this control uses common context and each formInfo block changes state of this context
2. comparison of boolean values using `myConst === true` doesn't make any sense and affect code readability
3. don't use misleading names of functions and variables, `toggleXXX` always means reversion of state, i.e. show if hidden and hide when visible

James-Baloyi:
Good afternoon, @IvanIlyichev - I've implemented the changes as requested

IvanIlyichev (changes requested):
Hi @James-Baloyi. Build fails

IvanIlyichev (approved):

Issue #2284, opened 2024-10-11
Allow dynamic menu items behavior to be configured


ihouvet:
### Overview
It is possible to configure dynamic in new items however, Both the list of items and the behavior upon clicking Requires coding. ideally this should be more configurable.

### Acceptance Criteria
1. It should be possible to define the list of the items to be displayed through configuration:
   1. this could include sourcing the list of items from:
      1. Entity + normal filter configuration
      2. An API
   2. It should then also be possible to specify which properties to use for the menu item caption and identifier (defaulted to Id if Entity)
2. It should be possible to define the behavior upon clicking a dynamic menu item through configuration in a similar way that you would for a normal menu item
   1. to enable this it should be possible to reference the selected item identifier and caption through variables.

PR #2352, opened 2024-10-31
Enhancement/dynamic-menu-items

Tshepo1103-lab:
#2284: Refactor Button Group for Configurable Dynamic Items

This update enhances the functionality of the button group component by making it more configurable and reducing the need for custom code.

Key Changes:
- Dynamic Item Retrieval: Allows users to specify a URL or entity type to dynamically fetch button group items.
- Flexible Filtering: Users can now define filtering criteria directly, making it easier to control displayed items based on specific conditions.
- This implementation empowers developers to configure dynamic button groups with minimal effort, improving adaptability and simplifying integration.


IvanIlyichev (changes requested):
Hi @Tshepo1103-lab (cc: @ihouvet)
It's breaking change. We already use dynamic items with custom providers (e.g. in enterprise project), after merge of this PR enterprise forms will be broken
Previous implementation of dynamic items support custom parameters. Functionality from this PR can be implementaed as a custom dynamic items provider, it allows to keep existing providers.

Issues with `Label Property` and `Tooltip Property`:
1. If you decided to use poroperty names instead of text values there should be an autocomplete instead of text editor
2. JS mode of these properties doesn't work even when I try to return a name of real property. I.e. when I bind to Person and use code `return 'firstName';` it doesn't work.
3. Usage of property names instead of text is not flexible solution. In current implementation the user is unable to use calculated values. It would be more flexible if the user can use JS expression to calculate text using code editor (with intellisense support). If we expect that binding to property will be the most common usage then we can add two modes: property (with property autocomplete) and text (with JS support).

General note: JS mode of all item settings should allow to use data of the item fetched from the back-end (with intellisense when possible). I see you pass `dynamicItem` to the action configuration context but intellisense in the code editor is not available

Issue #2234, opened 2024-10-03
Collapse form builder panels by default in preview mode


AlexanderSlavchovUX:
When you preview a form the current behaviour of the panels is that they remain on screen. You will see in the image attached below that the properties panel displays nothing and the widget panel still looks like it can be interacted with, although it too does not allow you to do anything with it. 

Can we please collapse these containers by default when a user clicks preview mode then restore them to the users set widths when they leave preview mode. Currently these panels serve no purpose other than cluttering the preview mode and misleading users that a widget can be interacted with. 

![Panels in Preview](https://github.com/user-attachments/assets/e59c54cc-3a30-4509-8505-d39c1c631cf4)

PR #2371, opened 2024-11-06
chore: Collapse form builder panels by default in preview mode

teboho:
implrements issue https://github.com/shesha-io/shesha-framework/issues/2234

to Collapse form builder panels by default in preview mode and bring them back when back in edit mode

- tested locally with where SidePanel react component is being used: `Datatable to edit columns`, `form designer`, `Entity Configurator`

https://github.com/user-attachments/assets/ce19b971-e6c8-4d9f-8d38-906ae1e873ee


IvanIlyichev (changes requested):
Hi @teboho. This logic shouldn't be a part of the SidebarContainer, it's a generic control and it's used in different places. Please move code for hiding panels to the upper level.
Also I think that it's better to hide panels in the preview mode rather then collapse them

teboho:
> Hi @teboho. This logic shouldn't be a part of the SidebarContainer, it's a generic control and it's used in different places. Please move code for hiding panels to the upper level.
> 
> Also I think that it's better to hide panels in the preview mode rather then collapse them

Thank for review @IvanIlyichev will look into it

teboho:
> Hi @teboho. This logic shouldn't be a part of the SidebarContainer, it's a generic control and it's used in different places. Please move code for hiding panels to the upper level. Also I think that it's better to hide panels in the preview mode rather then collapse them

Hi @IvanIlyichev. Thank you for the suggested changes, I modified the code accordingly:  
- Reverted the earlier commits to bring back the code to original state  
- Took the collapse control to the designerMainArea  
- - Now rendering the form outside of the sidebarContainer if in preview mode (```formMode !== 'designer'```) otherwise use the sidebarContainer like before if in 'designer' mode  
- Added a border top for preview mode to separate the rendered preview of the form from the designer toolbar and the rest of the upper section  

https://github.com/user-attachments/assets/fb04fca7-5304-476f-806d-be2f1402185e

IvanIlyichev (changes requested):
Hi @teboho. You created two copies of the same markup. Please minimize code duplication to avoid mistakes during code support (for instance wrap SidebarContainer with a ConditionalWrap).

teboho:
> Hi @teboho. You created two copies of the same markup. Please minimize code duplication to avoid mistakes during code support (for instance wrap SidebarContainer with a ConditionalWrap).

Hey @IvanIlyichev, I've made use of the ConditionalWrap please review. This was great feedback, finally getting to understand why there was a need for conditional wrap react component.

IvanIlyichev (approved):

PR #2372, opened 2024-11-06
Fix/dynamic menu

Tshepo1103-lab:
The implementation of the url and entity dynamic actions is only visible on shesha.

This is to make it visible in other projects. Wrapped the provider within the `ShaApplicationProvider` .


Tshepo1103-lab:
Related to this issue: 
#2284

IvanIlyichev (approved):

Issue #699, opened 2023-12-01
Formatting of nested panels should be cleaned-up by making smaller


Tshepiso-R:
### Description
Sometimes we have to nest panels within each other. Formatting is of nested panels is identical to top level panels which visually can be confusing and make the UI look busy/crowded.

Styling of nested panels should be changed to reinforce the visual hierarchy.

![image](https://github.com/shesha-io/shesha-framework/assets/151129946/c0f49d0f-bab8-49be-9ac4-e97cb402be91)

### Acceptance Criteria

1. Make the font size of the nested panel headers smaller and reduce vertical padding a bit
2. Perhaps make the font color a touch lighter (e.g. a dark grey instead of straight black)
3. Make blue separator bar thinner for nested panels

nthangeniphumudzo:
@hlayisanimazuze this seem to be resolved here is the [link](https://function-adminportal-test.shesha.dev/shesha/forms-designer?id=a74a6511-b575-4da5-839d-6b3643c61453). You can send the link of the configure where it is not working.

hlayisanimazuze:
@nthangeniphumudzo, correct the nesting of the panels is fixed, however there are additional requirements requested when the nesting panels is applied. Please check those requirements

![image](https://github.com/shesha-io/shesha-framework/assets/151041759/ca9d1a59-25ff-4b34-aaba-ac3ccab77e31)

AlexanderSlavchovUX:
Hi @nthangeniphumudzo 

Please see the mockup below. 

The changes here are as follows: 

- Panel header heights have been reduced to 60px
- Panel titles are in bold
- Nested panels have colour border on the left
- Nested panels header background is 5% opacity of the primary theme colour.

![Nested 1](https://github.com/shesha-io/shesha-framework/assets/153498650/8b4b02a0-3e48-4431-ac3b-972866d97c51)
![Nested 2](https://github.com/shesha-io/shesha-framework/assets/153498650/75a58638-10c7-48e6-b4ec-13630d03c465)
![Nested 3](https://github.com/shesha-io/shesha-framework/assets/153498650/c11709d8-c1a0-4543-b03f-e5ca2f2c8fd6)

AlexanderSlavchovUX:
Hi @nthangeniphumudzo  I'm updating the card after a conversation with @hlayisanimazuze regarding the space between nested items.

Here is an image provided by Hlayi on how it looks when there are many nested items
[...]

PR #2413, opened 2024-11-18
configurable header and primary borders

nthangeniphumudzo:
Hi @AlexStepantsov . This PR addressing the issue raised [here](https://github.com/shesha-io/shesha-framework/issues/699). I have ensured that the parent panel(outer) has a border-top with primary-color and the subsequent children to have a border-left with primary color. Finally I have added the custom Header option to allow user flexibility if the don't like a default header's layout.

### DIFF
--- a/shesha-reactjs/src/designer-components/collapsiblePanel/collapsiblePanelComponent.tsx
+++ b/shesha-reactjs/src/designer-components/collapsiblePanel/collapsiblePanelComponent.tsx
@@ -54,19 +54,28 @@ const CollapsiblePanelComponent: IToolboxComponent<ICollapsiblePanelComponentPro
     };
 
     const headerComponents = model?.header?.components ?? [];
+
+    const hasCustomHeader = model?.hasCustomHeader;
+
     const extra =
-      headerComponents?.length > 0 || formMode === 'designer' ? (
+      ((headerComponents?.length > 0 || formMode === 'designer') && !hasCustomHeader) ? (
         <ComponentsContainer
           containerId={model.header?.id}
           direction="horizontal"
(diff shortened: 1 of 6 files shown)


shesha-reactjs/src/designer-components/collapsiblePanel/collapsiblePanelComponent.tsx
```diff
@@ -54,39 +60,54 @@ const CollapsiblePanelComponent: IToolboxComponent<ICollapsiblePanelComponentPro
...
+            dynamicBorderRadius={model?.borderRadius}
+            style={{ ...getPanelStyle }}
+            className={model.className}
+            bodyColor={bodyColor}
+            headerColor={headerColor}
+            isSimpleDesign={isSimpleDesign}
+            parentPanel={panelPosition === 'parent'}
+            hideCollapseContent={hideCollapseContent}
+            hideWhenEmpty={hideWhenEmpty}
+            isSettingPanel={!model?.componentName}
```
AlexStepantsov: Don't ever do that!

It is completely unclear how the is Setting Panel property is related to Component Name. The configurator can specify Component Name and then your logic will break. You can't make such non-obvious relations!

To determine that the form is in settings mode, you can do the following:
`const form = useForm();`
...
`isSettingPanel={form.formSettings.isSettingsForm}`

In addition, it is completely wrong to add a form-specific settings mode to CollapsiblePanel, since this is a common component and should not be associated with the Form Designer in any way.

The correct solution would be to add some property, for example, Default Header Border or something, which will set the default header type.

Or сustomHeaderBorder with the ability to specify a specific style

AlexStepantsov (changes requested):

Issue #2307, opened 2024-10-20
Items on Button group and tab configuration panels should be easier to select


ihouvet:
### Overview
Currently on the button group configurator to select an item you have to click on the handle at the left of the item. This can be slightly irritating as you should be able to select the item by clicking on anywhere on the item bar

### Acceptance criteria
1. Allow an item to be selected by clicking anywhere on the item bar. This should work with any component that gets configured in this manner including:
   1. Button groups
   2. Main menu
   3. Tab component
   4. Entity configurator properties list
<img width="719" alt="image" src="https://github.com/user-attachments/assets/14e5b362-1c84-4f9f-a1c3-18513afdd482">

hlayisanimazuze:
Related item: https://github.com/shesha-io/shesha-framework/issues/2261

hlayisanimazuze:
PR

hlayisanimazuze:
Hi @James-Baloyi , The pull request for this item has broken the initial implementation related to items. Currently, all items inheriting this update are not functioning as expected:

- Delete Icon: When hovering over an item, there was originally a delete icon displayed. Now, the icon is missing, making it difficult to locate and use the delete function.
- ButtonGroup: When a group contains items, those items are now unselectable.

Rather than creating a bug report, let's review the implementation of this pull request and address the issues accordingly: [GitHub PR #2366](https://github.com/shesha-io/shesha-framework/pull/2366)."

hlayisanimazuze:
PR

hlayisanimazuze:
Required before creating a release branch

PR #2421, opened 2024-11-21
james/en/2307

James-Baloyi:
#2307 

There were issues with spacing between the drag handler and selected icon, and the button component capturing click and drag events.

### DIFF
--- a/shesha-reactjs/src/components/buttonGroupConfigurator/buttonGroupItem.tsx
+++ b/shesha-reactjs/src/components/buttonGroupConfigurator/buttonGroupItem.tsx
@@ -66,7 +66,7 @@ export const ButtonGroupItem: FC<IButtonGroupItemProps> = ({ item, actualModelCo
             size={size}
             block={block}
             disabled={readOnly}
-            style={{ ...getStyle(itemStyle), ...newStyles }}
+            style={{ ...getStyle(itemStyle), ...newStyles, pointerEvents: "none" }}
           >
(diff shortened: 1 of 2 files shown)


IvanIlyichev (changes requested):
Hi @James-Baloyi. Drag handle must not go beyond the current element (see attached image). Please review styles of the list editor, it's related not only to this PR, it's a results of your previous changes. 
image

James-Baloyi:
@IvanIlyichev the issue has been addressed

IvanIlyichev (changes requested):
Hi @James-Baloyi. The markup is still wrong. You are trying to align items using relative values (%) and absolute (px) at the same time. If we still use drag indicator it should be centered properly and dragging area shouldn't exceed the container. 
dragHandle - if you use absolute positioning, then set size using left/right/top/bottom properties
dragIcon cannot be centered using absolute values

IvanIlyichev (changes requested):
@James-Baloyi. Now **sha-drag-handle** overlaps **sha-list-item-controls**

James-Baloyi:
@IvanIlyichev I've reduced its footprint so it doesn't overlap with the controls, however, it previously was aligned below them so it would not affect their functionality.

IvanIlyichev:
@James-Baloyi yes, the alignment was incorrect even before this PR

IvanIlyichev (approved):

PR #2437, opened 2024-11-22
Charts: enhancements and code optimisations

teboho:
### Chart component enhancements and implicit bugfixes 

- wrap the component in ConfigurableFormItem
- conditional wrap if the user is allowed to filter data directly on the chart
- generating titles in a hook to abstract away the code and reduce duplication while keeping chart type components simple
- refined settings to handle 'hidden' situations
- width that works interchangeably with width to preserve chart responsiveness
- util function to stringify values on render to avoid chart js errors with reflists and undefined values
- enable pie charts to turn into doughnuts

*Some screenshots*

filtering on the fly

left legend

Doughnut


IvanIlyichev (changes requested):
Hi @teboho. Build fails, make sure that your code is compilable

teboho:
Alright I'm making further changes I reopen the PR on Monday

Issue #2311, opened 2024-10-21
Add user.personId front-end API to get the current user's person id and utilize it in the Js and filter configurator


ihouvet:
### Overview
The front end currently has a user api which allows you to return various properties of a user but it does not seem to expose a person ID which is often useful especially in filtering. see: [Javascript API docs] (https://dev.azure.com/boxfusion/Shesha%20Web%20v3.0/_wiki/wikis/Shesha-Web-v3.0.wiki/990/JavaScript-API)

**Describe the solution you'd like**
### Acceptance Criteria
1. A `user.personId` API should be available from any Js editor (incl. on function evaluate on filter configurator)
2. Documentation should be updated to reflect this

teboho:
function evaluate on filter configurator does not have direct access to the `application`  object so it can get acess to this via the formData

hlayisanimazuze:
I've created a separate card to handle the user object to be supported on query builder and mustache expression

PR #2448, opened 2024-11-25
Add user.personId front-end API to get the current user's person id

teboho:
Add user.personId front-end API to get the current user's person id and utilize it in the code editor wherever `application` is exposed


teboho:
Related issue https://github.com/shesha-io/shesha-framework/issues/2311

IvanIlyichev (approved):

Issue #1632, opened 2024-06-05
PostgreSQL Frwk_ConfigurationItems triggers


AlexStepantsov:
**Describe the bug**
There are errors on bulk insert/update in the Frwk_ConfigurationItems table 

**To Reproduce**
Migrations M20240501134700, M20240501141100 failed due to trigger errors

ERROR: the subquery returned more than one row
CONTEXT: SQL statement "(select "ModuleId" from inserted)"
PL/pgSQL function "log_Frwk_ConfigurationItems_UpdateIsLast_AI"(), line 7, RAISE statement

It seems that triggers are implemented for inserting/updating a single row, but errors appear when inserting in bulk.

In addition, the triggers differ from similar triggers in the MS SQL database (for example, there are no conditions for TenantId and ApplicationId)

**Expected behavior**
migrations done

**Need to create a migration with the triggers fixed and assign it a number before M20240501134700 so that migrations M20240501134700, M20240501141100 can done correctly**

hlayisanimazuze:
I've closed this: https://github.com/shesha-io/shesha-framework/issues/1531, I will retest it when the main card is fixed

IvanIlyichev:
Hi @AlexStepantsov. I've made changes on my side, it's already merged to main. Please review all changes and PR made by @teboho to ensure that all fixed properly

hlayisanimazuze:
Hi @IvanIlyichev , there are some triggers @teboho still needs to update based on the additional information @AlexStepantsov provided above. We will update the item once everything is satisfied.

teboho:
The conditions in the triggers cannot be updated yet until the .backup file for PostgreSQL is updated to have the same starter data as the .bacpac for SQL Server...

teboho:
The current state of the card IIRC, testing is required, a dev needs to run shesha-functional-tests on SQL Server, and also at the same time run another instance of shesha-functional-tests backend on PostgreSQL and then compare all the triggers on both DBs to make sure the ones on PostgreSQL behave exactly like the ones on SQL Server

PR #2477, opened 2024-11-27
fix: first load  initial migrations running on PostgreSQL

teboho:


teboho:
related issue https://github.com/shesha-io/shesha-framework/issues/1632

AlexStepantsov (approved):

PR #2497, opened 2024-12-04
Email and Sms Notifications rework V2

OmolemoBlessingLethuloe:


shesha-core/src/Shesha.Application/Notifications/NotificationAppService.cs
```diff
@@ -1,536 +1,336 @@
...
-    /// <param name="notificationName">Name of the notification. Default email template of the specified notification will be used</param>
-    /// <param name="data">Data that is used to fill template</param>
-    /// <param name="recipient">Receiptient of the notification</param>
-    /// <param name="sourceEntity">Optional parameter. If notification is an Entity level notification, specifies the entity the notification relates to.</param>
-    /// <returns></returns>
-    public async Task PublishSmsNotificationAsync<TData>(string notificationName,
-        TData data,
-        Person recipient,
-        GenericEntityReference sourceEntity = null) where TData : NotificationData
+    public class NotificationAppService: SheshaAppServiceBase
```
IvanIlyichev: General comment for all methods: please move all complex logic to the domain service/manager and call it from this application service. Application service should be used by the front-ends only, all internal code will call domain service/manager

shesha-core/src/Shesha.Application/Notifications/INotificationChannelSender.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Net.Mail;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Shesha.Notifications
+{
+    public interface INotificationChannelSender
+    {
+        string GetRecipientId(Person person);
+        Task<Tuple<bool, string>> SendAsync(Person fromPerson, Person toPerson, NotificationMessage message, bool isBodyHtml, string cc, List<EmailAttachment> attachments = null);
```
IvanIlyichev: Please replace usage of tuples with a new response class. Code is more readable when you use something like { isSuccess: bool, errorMessage: string } instead of tuple. We'll have to replace it in any case if we'll need any additional property to the response

shesha-core/src/Shesha.Application/Sms/SmsSettingsAppService.cs
```diff
@@ -82,11 +82,11 @@ public async Task UpdateSettingsV2(SmsSettingsV2Dto input)
                 ? _typeFinder.Find(t => typeof(ISmsGateway).IsAssignableFrom(t) && SmsUtils.GetGatewayAlias(t) == input.Gateway).FirstOrDefault()
                 : null;
 
-            var gatewayUid = gatewayType?.GetClassUid();
+            var gatewayName = gatewayType?.GetDisplayName();
```
IvanIlyichev: Please remove this class (SmsSettingsAppService). It's not used anymore

shesha-core/src/Shesha.Application/Notifications/Sms/SmsChannelSender.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            var settings = await GetSettings();
+
+            if (!settings.IsSmsEnabled)
+            {
+                Logger.Warn("SMSs are disabled");
+                return await Task.FromResult(new Tuple<bool, string>(false, "SMSs are disabled."));
+            }
+
+            var sender = StaticContext.IocManager.Resolve<ISmsGateway>();
```
IvanIlyichev: Please avoid usage of StaticContext, all dependencies should be retrieved using constructor injection or property injection where possible. StaticContext is very specific thing and created for cases where normal injection is not available

shesha-core/src/Shesha.Application/Notifications/Jobs/BroadcastNotificationJobQueuer.cs
```diff
@@ -0,0 +1,147 @@
...
+
+            var channel = await _notificationChannelRepository.GetAsync(args.ChannelId);
+
+            var message = await SaveUserNotificationMessagesAsync(notification, template, channel, args.Subject, args.Message, args.Attachments);
+
+            var senderChannelInterface = _channelSenders.FirstOrDefault(x => x.GetType().Name == channel.SenderTypeName);
+            if (senderChannelInterface == null)
+                throw new Exception($"No sender found for sender type: {channel.SenderTypeName}");
+
+            var sender = StaticContext.IocManager.Resolve<NotificationSender>(senderChannelInterface);
```
IvanIlyichev: 1. Please avoid usage of StaticContext, all dependencies should be retrieved using constructor injection or property injection where possible. StaticContext is very specific thing and created for cases where normal injection is not available
2. Use `INotificationSender` interface instead of class

shesha-core/src/Shesha.Application/Notifications/Jobs/DirectNotificationJobQueuer.cs
```diff
@@ -0,0 +1,46 @@
...
+        public override async Task ExecuteAsync(DirectNotificationJobArgs args)
+        {
+            var fromPerson = await _personRepository.GetAsync(args.FromPerson);
+            var toPerson = await _personRepository.GetAsync(args.ToPerson);
+            var notificationMessage = await _notificationMessage.GetAsync(args.Message);
+            var senderChannelInterface = _channelSenders.FirstOrDefault(x => x.GetType().Name == args.SenderTypeName);
+            if (senderChannelInterface == null)
+                throw new Exception($"No sender found for sender type: {args.SenderTypeName}");
+
+            var sender = StaticContext.IocManager.Resolve<NotificationSender>(senderChannelInterface);
```
IvanIlyichev: 1. Please avoid usage of StaticContext, all dependencies should be retrieved using constructor injection or property injection where possible. StaticContext is very specific thing and created for cases where normal injection is not available
2. Use `INotificationSender` interface instead of class

shesha-core/src/Shesha.Application/Notifications/NotificationAppService.cs
```diff
@@ -1,536 +1,336 @@
...
-                templateId: notificationTemplate,
-                data: data,
-                attachments: attachments,
-                emailAddress: emailAddress,
-                sourceEntity: sourceEntity,
-                cc: cc);
-
-            messageId = _notificationPublicationContext.Statistics.NotificationMessages.FirstOrDefault()?.Id;
-        }
+            var sender = StaticContext.IocManager.Resolve<NotificationSender>(senderChannelInterface);
```
IvanIlyichev: 1. Please avoid usage of StaticContext, all dependencies should be retrieved using constructor injection or property injection where possible. StaticContext is very specific thing and created for cases where normal injection is not available
2. Use `INotificationSender` interface instead of class

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        [UnitOfWork]
+        public async Task SendAsync(Person fromPerson, Person toPerson, NotificationMessage message, bool isBodyHtml)
+        {
+            var attachments = await GetAttachmentsAsync(message);
+            int attempt = 0;
+            Tuple<bool, string> sendStatus = new Tuple<bool, string>(false, "");
+
+            while (attempt < MaxRetries && !sendStatus.Item1)
```
IvanIlyichev: Retries should be handled by scheduled jobs and each attempts must be created using different DB sessions, otherwise any temporary DB issue will lead to a failed sending

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,200 @@
...
+        [UnitOfWork]
+        public async Task SendBroadcastAsync(Notification notification, string subject, string messageContent, List<EmailAttachment> attachments)
+        {
+            int attempt = 0;
+            Tuple<bool, string> sendStatus = new Tuple<bool, string>(false, "");
+
+            // Get all notification messages associated with the notification
+            var messages = _notificationMessageRepository.GetAll().Where(m => m.PartOf.Id == notification.Id).ToList();
+
+            while (attempt < MaxRetries && !sendStatus.Item1)
```
IvanIlyichev: Retries should be handled by scheduled jobs and each attempts must be created using different DB sessions, otherwise any temporary DB issue will lead to a failed sending

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,200 @@
...
+                // Increment retry count and save to the database
+                attempt++;
+                message.RetryCount = attempt;
+                message.ErrorMessage = $"Failed to send notification after {attempt} attempts. Message: {sendStatus.Item2}";
+                message.Status = RefListNotificationStatus.Failed;
+
+                using (var uow = _unitOfWorkManager.Begin())
+                {
+
+                    using (var transaction = _sessionProvider.Session.BeginTransaction())
```
IvanIlyichev: See my comment above, single attempt should be atomic and work as a single unit of work:

- read required data
- try to send
- save status
- schedule next attempt if failed

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,200 @@
...
+                            Logger.ErrorFormat($"Attempt {attempt + 1} to send notification failed. Message: {sendStatus.Item2}");
+                            message.Status = RefListNotificationStatus.Failed;
+                            message.RetryCount = attempt;
+                            message.ErrorMessage = $"Failed to send notification on attempt {attempt + 1}. Message: {sendStatus.Item2}";
+                        }
+
+                        // Save changes to each message
+                        using (var uow = _unitOfWorkManager.Begin())
+                        {
+                            using (var transaction = _sessionProvider.Session.BeginTransaction())
```
IvanIlyichev: See my comment above, single attempt should be atomic and work as a single unit of work:

- read required data
- try to send
- save status
- schedule next attempt if failed

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,200 @@
...
+                catch (Exception ex)
+                {
+                    Logger.ErrorFormat($"Exception on attempt {attempt + 1} to send broadcast notification: {ex.Message}");
+                }
+
+                // Introduce a delay before retrying
+                attempt++;
+                if (attempt < MaxRetries)
+                {
+                    await Task.Delay(1000); // Backoff delay
```
IvanIlyichev: This code keep current connection opened and may lead to a DB locks, it's additional reason why sending should be atomic

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,200 @@
...
+                        await _notificationMessageRepository.UpdateAsync(message);
+                        transaction.Commit();
+                    }
+                    await uow.CompleteAsync();
+                }
+
+                // Introduce a delay before the next retry attempt
+                if (attempt < MaxRetries)
+                {
+                    await Task.Delay(1000); // Backoff delay
```
IvanIlyichev: This code keep current connection opened and may lead to a DB locks, it's additional reason why sending should be atomic

shesha-core/src/Shesha.Application/SheshaApplicationModule.cs
```diff
@@ -50,10 +56,7 @@ public override void PreInitialize()
             Configuration.Auditing.IsEnabled = false;
 
             IocManager.Register<IShaApplicationModuleConfiguration, ShaApplicationModuleConfiguration>();
-
-            Configuration.Notifications.Providers.Add<ShaNotificationProvider>();
-            Configuration.Notifications.Notifiers.Add<EmailRealTimeNotifier>();
-            Configuration.Notifications.Notifiers.Add<SmsRealTimeNotifier>();
+            IocManager.Register<INotificationSender, NotificationSender>();
```
IvanIlyichev: Please use conventional registration instead of manual where possible

shesha-core/src/Shesha.Application/Notifications/Configuration/NotificationSettings.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    /// <summary>
+    /// Enter preferred channels per priority.
+    /// </summary>
+    public class NotificationSettings
+    {
+        /// <summary>
+        /// Low priority notifications
+        /// </summary>
+        public List<NotificationChannelConfig> Low { get; set; }
```
IvanIlyichev: `NotificationChannelConfig` is an entity. I must not be saved as part of settings

shesha-core/src/Shesha.Application/Notifications/NotificationAppService.cs
```diff
@@ -1,536 +1,336 @@
...
-    //    return entityIdentifier;
-    //}
-
-    private static EntityIdentifier GetEntityIdentifier(GenericEntityReference genericEntity)
-    {
-        EntityIdentifier entityIdentifier = null;
+            if (channel != null)
+            {
+                var template = await _messageTemplateRepository.FirstOrDefaultAsync(x => x.PartOf.Id == type.Id && channel.SupportedFormat == x.MessageFormat);
+                var subject = TemplateHelper.ReplacePlaceholders(template.TitleTemplate, data);
```
IvanIlyichev: template may be null here

shesha-core/src/Shesha.Application/Notifications/NotificationAppService.cs
```diff
@@ -1,536 +1,336 @@
...
+                if (subscriptions != null && subscriptions.Any())
+                {
+                    foreach (var user in users)
+                    {
+                        var userChannels = await GetChannelsAsync(type, user.User, priority);
+
+                        foreach (var channelConfig in userChannels)
+                        {
+                            var template = await _messageTemplateRepository.FirstOrDefaultAsync(x => x.PartOf.Id == type.Id && channelConfig.SupportedFormat == x.MessageFormat);
+                            var subject = TemplateHelper.ReplacePlaceholders(template.TitleTemplate, data);
```
IvanIlyichev: template may be null here

shesha-core/src/Shesha.Application/Notifications/NotificationAppService.cs
```diff
@@ -1,536 +1,336 @@
...
-            else
+
+            // Step 2: Check Override Channels in NotificationTypeConfig
+            if (!string.IsNullOrEmpty(type.OverrideChannels))
             {
-                await SendNotificationByType(notificationName, (int)RefListNotificationType.SMS, recipient, data, sourceEntity, attachments);
-                return await SendNotificationByType(notificationName, (int)RefListNotificationType.Email, recipient, data, sourceEntity, attachments, cc);
+                try
+                {
+                    var overrideChannelIds = JsonSerializer.Deserialize<Guid[]>(type.OverrideChannels);
```
IvanIlyichev: Please move login of serialization/deserialization to the NotificationTypeConfig (add a new property with getter and setter). Also it's better to save composite identifier instead of guid. NotificationChannelConfig is a ConfigurationItem, so we can identify it using pair modulename+name (check usage of FormIdentifier as a reference)

shesha-core/src/Shesha.Application/Notifications/NotificationAppService.cs
```diff
@@ -1,536 +1,336 @@
...
+                        .ToListAsync();
+
+                    if (channels.Any())
+                    {
+                        return channels;
+                    }
+                }
+                catch (JsonException ex)
+                {
+                    new UserFriendlyException("Error deserializing override channels", ex);
```
IvanIlyichev: You create a new exception but don't throw it, it affects entire the process of notifications sending.

shesha-core/src/Shesha.Application/Notifications/INotificationChannelSender.cs
IvanIlyichev: Please check usage of `isHtmlBody`, looks like it was copied from email-specific functionality. This flag should be a part of the template

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,200 @@
...
+
+            // Get all notification messages associated with the notification
+            var messages = _notificationMessageRepository.GetAll().Where(m => m.PartOf.Id == notification.Id).ToList();
+
+            while (attempt < MaxRetries && !sendStatus.Item1)
+            {
+                try
+                {
+                    // Attempt to send the message via the channel sender
+                    sendStatus = await _channelSender.BroadcastAsync(notification.NotificationTopic, subject, messageContent, attachments);
```
IvanIlyichev: Result of this call and possible exceptions are not handled properly. If exception will be thrown during the message preparation you will not know it and status of messages will not be updated.

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,200 @@
...
+
+
+        [UnitOfWork]
+        public async Task SendBroadcastAsync(Notification notification, string subject, string messageContent, List<EmailAttachment> attachments)
+        {
+            int attempt = 0;
+            Tuple<bool, string> sendStatus = new Tuple<bool, string>(false, "");
+
+            // Get all notification messages associated with the notification
+            var messages = _notificationMessageRepository.GetAll().Where(m => m.PartOf.Id == notification.Id).ToList();
```
IvanIlyichev: Please review loading of **NotificationMessage**s. They are loaded here, but used only after sending, also **_channelSender** loads them again when it prepares a list of email addresses. There are two options:
1. NotificationSender loads all data itself and then passes as arguments to the _channelSender, then updates statuses after sending attempt
2. _channelSender loads notifications itself and update statuses (maybe using callbacks/events)
Also I think we'll need a **resend** functionality, in this case we'll have to skip notification messages which were already successfully sent on a previous attempt and this filtration should be implemented in one place and not copied to both **NotificationSender** and **_channelSender**.

shesha-core/src/Shesha.Application/Notifications/NotificationAppService.cs
```diff
@@ -1,536 +1,336 @@
...
-        }
+            if (senderChannelInterface == null)
+                throw new UserFriendlyException($"Sender not found for channel {channelConfig.Name}");
 
-        return messageId;
-    }
+            // Create a new notification message
+            var message = await SaveOrUpdateEntityAsync<NotificationMessage>(null, item =>
+            {
+                item.PartOf = notification;
```
IvanIlyichev: Looks like something is wrong with the **NotificationMessage** usage. I found only this line where you set a value of the **PartOf** property. Also other properties are not used at all, please check **RecipientText** and **DateSent**

shesha-core/src/Shesha.Application/Notifications/Emails/EmailChannelSender.cs
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        public string GetRecipientId(Person person)
+        {
+            return person.EmailAddress1;
+        }
+
+        public async Task<string> GetRecipients(NotificationTopic topic)
+        {
+            var recipients = await _userTopicSubscriptionRepository.GetAll().Where(s => s.Topic.Id == topic.Id).Select(s => GetRecipientId(s.User)).ToListAsync();
```
IvanIlyichev: Make sure that this code works. **GetRecipientId** cannot be translated to the SQL, I believe it will fail.

IvanIlyichev (changes requested):
Hi @OmolemoBlessingLethuloe . Please review and resolve all comments, then I'll be able to continue the review

shesha-core/src/Shesha.Application/SheshaApplicationModule.cs
IvanIlyichev: **INotificationSender** is still registered manually. Please remove this registration and add **ITransientDependency** to the **NotificationSender**

shesha-core/src/Shesha.Application/Notifications/Configuration/NotificationSettings.cs
IvanIlyichev: 1. Notifications Channels setting doesn't work. Try to changes settings, save and refresh refresh page. Maybe it's a combination of storage mechanism and form configuration. 
2. I checked, you changed storage of entities directly in the settings to guids usage. **NotificationChannelConfig** is a configuration item and it's identified by a pair **module+name**, so it's more correct to save module and name as JSON instead of guid. Guid is pointing to a concrete version of configuration item, current implementation may lead to usage of outdated configurations, or as an alternative you'll have to search latest/live version using non-standard approach (fetch by id, analyze module+name, search for the latest/live version), in any case it will work less stable

shesha-core/src/Shesha.Application/Notifications/NotificationAppService.cs
IvanIlyichev: Usage of **FirstOrDefault()** here may lead to unexpected results. Domain model allows to create more than one template for the same notification with the same message format. Something is wrong: domain model or the logic. I see two options:
1. disallow to creation more than one templates with the same parameters using unique constraints and validation
2. handle it in the code properly: process all templates/choose one of them according to some additional conditions

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            try
+            {
+                var sendStatus = await notificationChannelSender.SendAsync(fromPerson, toPerson, message, "", attachments);
+
+                // Update the message status based on send result
+                if (sendStatus.IsSuccess)
+                {
+                    message.Status = RefListNotificationStatus.Sent;
+                    message.RetryCount = 0; // No retry as this is the first (successful) attempt
```
IvanIlyichev: We shouldn't clear number of retries after successful sending, this information may be useful for audit purposes

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,145 @@
...
+                    message.ErrorMessage = null; // Clear any previous error
+                    message.RecipientText = notificationChannelSender.GetRecipientId(toPerson);
```
IvanIlyichev: RecipientText should be set on both successful/failed attempt. It's important because the RecipientId (email/mobile/etc) may be different on different attempts

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            var attachments = await GetAttachmentsAsync(message);
+
+            try
+            {
+                var sendStatus = await notificationChannelSender.SendAsync(fromPerson, toPerson, message, "", attachments);
+
+                // Update the message status based on send result
+                if (sendStatus.IsSuccess)
+                {
```
IvanIlyichev, line 247: After successful sending we should set **DateSent**

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,145 @@
...
+                    message.RetryCount = 0; // No retry as this is the first (successful) attempt
+                    message.ErrorMessage = null; // Clear any previous error
+                    message.RecipientText = notificationChannelSender.GetRecipientId(toPerson);
+
+                    await _notificationMessageRepository.UpdateAsync(message);
+                }
+                else
+                {
+                    // Increment the retry count for the message
+                    message.RetryCount++;
```
IvanIlyichev, line 258: This code is a duplicate of the catch block below and they are different. Please review them and remove the ambiguity, they must work by the same way, otherwise any investigation of failure reasons will be more complex

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <param name="notificationChannelSender"></param>
+        /// <returns></returns>
+        [UnitOfWork]
+        public async Task SendAsync(Person fromPerson, Person toPerson, NotificationMessage message, INotificationChannelSender notificationChannelSender)
+        {
+            var attachments = await GetAttachmentsAsync(message);
+
+            try
+            {
+                var sendStatus = await notificationChannelSender.SendAsync(fromPerson, toPerson, message, "", attachments);
```
IvanIlyichev: In case of failure you have two possible results: 
1. sendStatus with isSuccess = false
2. exception thrown
To make the handling of result easier you can create a private method **TrySendAsync()** as part of **NotificationSender**, it must not generate any exceptions and return a response. Using local method you will eliminate uncertainty and handling of result will be the same irrespectively of implementation of channel senders

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
IvanIlyichev: PS: see my comment about **TrySendAsync** above

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,145 @@
...
+                message.Status = RefListNotificationStatus.Failed;
+                message.ErrorMessage = $"Exception while sending notification: {ex.Message}";
+                message.RetryCount++;
+
+                await _notificationMessageRepository.UpdateAsync(message);
+
+                Logger.Error($"Exception while sending notification {message.Id}: {ex.Message}");
+
+                // Schedule a retry job if an exception occurs
+                ScheduleDirectRetryJob(message, notificationChannelSender);
```
IvanIlyichev: Next attempt should be scheduled with a delay, in this case we have more chances to send it successfully. Use Hangfire as a reference, see https://docs.hangfire.io/en/latest/tutorials/send-email.html#automatic-retries

shesha-core/src/Shesha.Application/Notifications/Jobs/DirectNotificationJobQueuer.cs
```diff
@@ -0,0 +1,112 @@
...
+
+            try
+            {
+                await _notificationSender.SendAsync(fromPerson, toPerson, notificationMessage, senderChannel);
+
+                Logger.Info($"Notification {notificationMessage.Id} sent successfully on attempt {attempt + 1}");
+            }
+            catch (Exception ex)
+            {
+                notificationMessage.RetryCount = attempt + 1;
```
IvanIlyichev: This code is a duplicate of **NotificationSender.SendAsync**. There should be a single method that handles successful and failed attempts. In fact this job class is not required and it can be moved to the NotificationSender. Background job class is required only for [recurrent tasks](https://docs.hangfire.io/en/latest/background-methods/performing-recurrent-tasks.html#performing-recurrent-tasks) which can be scheduled.

shesha-core/src/Shesha.Core/Domain/Enums/RefListChannelSupportedMechanism.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Shesha.Domain.Attributes;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Shesha.Domain.Enums
+{
+    [ReferenceList("Shesha.Core", "ChannelSupportedMechanism")]
```
IvanIlyichev: Please remove namespace and use constructor that accepts name only (in this case reflist is linked to module). I see this constructor is marked as obsolete only on base class, I'll update it

shesha-core/src/Shesha.Core/Domain/Enums/RefListNotificationChannelStatus.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Shesha.Domain.Attributes;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Shesha.Domain.Enums
+{
+    [ReferenceList("Shesha.Core", "NotificationChannelStatus")]
```
IvanIlyichev: Please remove namespace and use constructor that accepts name only (in this case reflist is linked to module). I see this constructor is marked as obsolete only on base class, I'll update it

shesha-core/src/Shesha.Core/Domain/Enums/RefListNotificationMessageFormat.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Shesha.Domain.Attributes;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Shesha.Domain.Enums
+{
+    [ReferenceList("Shesha.Core", "NotificationMessageFormat")]
```
IvanIlyichev: Please remove namespace and use constructor that accepts name only (in this case reflist is linked to module). I see this constructor is marked as obsolete only on base class, I'll update it

shesha-core/src/Shesha.Core/Domain/Enums/RefListNotificationPriority.cs
```diff
@@ -0,0 +1,18 @@
+﻿using Shesha.Domain.Attributes;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Shesha.Domain.Enums
+{
+    [ReferenceList("Shesha.Core", "NotificationPriority")]
```
IvanIlyichev: Please remove namespace and use constructor that accepts name only (in this case reflist is linked to module). I see this constructor is marked as obsolete only on base class, I'll update it

shesha-core/src/Shesha.Core/Domain/Enums/RefListNotificationReadStatus.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Shesha.Domain.Attributes;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Shesha.Domain.Enums
+{
+    [ReferenceList("Shesha.Core", "NotificationReadStatus")]
```
IvanIlyichev: Please remove namespace and use constructor that accepts name only (in this case reflist is linked to module). I see this constructor is marked as obsolete only on base class, I'll update it

shesha-core/src/Shesha.Core/Domain/Enums/RefListChannelSupportedMechanism.cs
IvanIlyichev: This list is used as multivalue, in this case it should use bit numbering and values should be 1,2,4,8 etc

shesha-core/src/Shesha.Core/Domain/Enums/RefListNotificationMessageFormat.cs
IvanIlyichev: if this list if used as multivalue it should use bit numbering as well

shesha-core/src/Shesha.Core/Domain/Enums/RefListNotificationPriority.cs
IvanIlyichev: if this list if used as multivalue it should use bit numbering as well

shesha-core/src/Shesha.Application/Notifications/NotificationAppService.cs
```diff
@@ -1,536 +1,190 @@
...
+        private readonly IRepository<NotificationTemplate, Guid> _messageTemplateRepository;
+        private readonly IRepository<StoredFile, Guid> _storedFileRepository;
+        private readonly INotificationSettings _notificationSettings;
+        private readonly IStoredFileService _storedFileService;
+        private readonly IBackgroundJobManager _backgroundJobManager;
+        private readonly INotificationSender _notificationSender;
+        private readonly INotificationManager _notificationManager;
+
+
+        public NotificationAppService(IEnumerable<INotificationChannelSender> channelSenders,
```
IvanIlyichev: This application service doesn't contain any endpoints, what is the purpose of this service then? Which endpoints can we use to send notifications from the front-end?

shesha-core/src/Shesha.Core/Migrations/M20241203150000.cs
```diff
@@ -0,0 +1,193 @@
...
+                  .WithColumn(DatabaseConsts.ExtSysId).AsString(50).Nullable()
+                  .WithColumn(DatabaseConsts.ExtSysLastSyncDate).AsDateTime().Nullable()
+                  .WithColumn(DatabaseConsts.ExtSysSource).AsString(50).Nullable()
+                  .WithColumn(DatabaseConsts.ExtSysSyncError).AsStringMax().Nullable()
+                  .WithColumn(DatabaseConsts.ExtSysSyncStatusLkp).AsInt32().Nullable();
+
+            Alter.Table("Core_NewNotificationMessages")
+                  .AddTenantIdColumnAsNullable();
+
+            Execute.Sql(@"
```
IvanIlyichev: Review scripts and make different updates for MS Sql and PostgreSql

shesha-core/src/Shesha.Core/Migrations/M20241203150000.cs
```diff
@@ -0,0 +1,193 @@
...
+	                Id,
+	                Subject,
+	                Body,
+	                BodyFormatLkp
+                FROM Core_NotificationTemplates
+                WHERE IsDeleted = 0;
+            ");
+
+
+            Rename.Table("Core_NotificationMessages").To("Core_OldNotificationMessages");
```
IvanIlyichev: Please rename tables before creation of new ones, otherwise some auto-generated names (FKs, indexes etc.) will contain wrong table names with `New` part

shesha-core/src/Shesha.Application/Notifications/Dto/DirectNotificationJobArgs.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Shesha.Domain;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Shesha.Notifications.Dto
+{
+    public class DirectNotificationJobArgs
```
IvanIlyichev: Please rename properties which contain identifiers to xxxId, e.g. FromPerson -> FromPerson**Id**

shesha-core/src/Shesha.Application/Notifications/Configuration/NotificationSettings.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Shesha.Notifications.Configuration
+{
+    /// <summary>
+    /// Enter preferred channels per priority.
+    /// </summary>
+    public class NotificationSettings
```
IvanIlyichev, line 16: 1. Notifications Channels setting doesn't work on the front-end. Try to changes settings, save and refresh refresh page. Maybe it's a combination of storage mechanism and form configuration.
2. **NotificationChannelConfig** is a configuration item and it's identified by a pair **module+name**, so it's more correct to save module and name as JSON instead of guid. Guid is pointing to a concrete version of configuration item, current implementation may lead to usage of outdated configurations, or as an alternative you'll have to search latest/live version using non-standard approach (fetch by id, analyze module+name, search for the latest/live version), in any case it will work less stable. Also we have some standard code that handles module+name and takes current selected mode (latest/ready/live) into account, check usage of **ConfigurationItemsExtensions**.

IvanIlyichev (changes requested):
Hi @OmolemoBlessingLethuloe. Please do the following:
1. Review all inline comments and resolve
2. Remove all console logging from the embedded forms
3. Rename embedded forms which are used for index pages to use plural names (e.g. notification-templates instead of notification-template-table)
4. **notification-type-config-table** form: **Update Channel Overrides** function doesn't work
Also please clarify purpose of the default NotificationTypeConfigs (Info/Warning/Emergency/OTP) imported automatically.

shesha-core/src/Shesha.Application/Notifications/NotificationManager.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            // Deserialize notification identifiers
+            var notificationIdentifiers = selectedNotifications
+                .Select(json => JsonSerializer.Deserialize<NotificationChannelIdentifier>(json, new JsonSerializerOptions
+                {
+                    PropertyNameCaseInsensitive = true
+                }))
+                .ToList();
+
+            var liveChannels = _notificationChannelRepository
```
IvanIlyichev, line 86: One of ConfigurationItemsExtensions methods should be used here. I'll review it myself after merge, this branch is out of date.

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -0,0 +1,305 @@
...
+        /// <param name="toPerson"></param>
+        /// <param name="type"></param>
+        /// <param name="priority"></param>
+        /// <param name="channelConfig"></param>
+        /// <param name="attachments"></param>
+        /// <returns></returns>
+        /// <exception cref="UserFriendlyException"></exception>
+        private async Task SendNotificationToChannel<TData>(Notification notification, TData data, Person fromPerson, Person toPerson, NotificationTypeConfig type, RefListNotificationPriority priority, NotificationChannelConfig channelConfig, List<NotificationAttachmentDto> attachments = null) where TData : NotificationData
+        {
+            var template = await _messageTemplateRepository.FirstOrDefaultAsync(x => x.PartOf.Id == type.Id && channelConfig.SupportedFormat == x.MessageFormat);
```
IvanIlyichev, line 171: I think I mentioned it somewhere but can't find my comment. Domain model allows to store multiple template with the same message format. This code will work by a wrong way when there are more than one template with the same format. Options to solve it:
1. restrict creation of multiple templates with the same format
2. handle all existing templates
3. use single template but sort the query by some columns
Note: in any case the logic should be transparent, please review the requirements an choose one of ways

IvanIlyichev (changes requested):
Hi @OmolemoBlessingLethuloe 
1. Please check inline comments
2. Regarding forms:
`shesha/notification-type-configs` - I think that  should be a normal index page without inline editing and it also should contain operations specific to all other configuration items (Publish, Set Ready, Create new version etc.). 
Also we need a details page (`shesha/notification-type-config-details`), it should include config details and a list of all templates of the notification.

'shesha/notification-type-config' - should be a normal index page with configuration items specific operations.

Settings: Notification Channels: doesn't save/show multiple selection, single items are displayed as JSON

4. Please remove unused packages:
- package20241204_1336.shaconfig
- package20241204_1346.shaconfig
- package20241209_1439.shaconfig 

5. We need an alternative to old API endpoint `/api/services/app/Notification/Publish` that allows to send a notificaiton

shesha-core/src/Shesha.Core/Migrations/M20241203150000.cs
```diff
@@ -0,0 +1,338 @@
...
+                                                      'Core_Notifications'))
+                        LOOP
+                            EXECUTE 'ALTER TABLE ' || quote_ident(r.table_schema) || '.' || quote_ident(r.table_name) 
+                                || ' DROP CONSTRAINT IF EXISTS ' || quote_ident(r.constraint_name);
+                        END LOOP;
+                    END $$;
+                ");
+
+                // For PostgreSQL, we also need to drop default values set using ALTER COLUMN
+                Execute.Sql(@"
```
IvanIlyichev: Migration fails here on MS Sql because this statement is executed without type of DBMS. Note: you needn't any complex sql for dropping/altering of constraints/indexes. Their names must be identical in any DB of the save version.

IvanIlyichev (changes requested):
Hi @OmolemoBlessingLethuloe. Please check inline comment

IvanIlyichev (changes requested):
Hi @OmolemoBlessingLethuloe. Migration **M20241203150000** fails on PostgreSql. 
Note: I faced with other migration issues on PostgreSql and fixed them, fixes merged to main. You can merge main to your branch, create new DB using backup from the repo and check your migration

IvanIlyichev (approved):

PR #2561, opened 2025-01-03
add action configurator onto icon picker

sadikiderik:
Add onClickEventHandler

- Should allow the user to configure and trigger a Dialog Modal upon clicking the icon.


sadikiderik:
@Tshepo1103-lab

PR #2605, opened 2025-01-20
en: tabbed properties for column

teboho:


teboho:
@James-Baloyi just pushed the gutter fallback commit

James-Baloyi (approved):

PR #2609, opened 2025-01-21
Fix notification framework

Tshepo1103-lab:
1. Fixed the migration to add attachments to `NotificationMessageAttachement`
2. Fixed the Spelling on `CanOptOut` property
3.  Added the `SupportAttachments` property on NotificationChannelConfig
4. Removed encoding for the body, due to decoding on the `Clickatell` not working.
5. Added Config Migrations after making changes to the forms
6. Added checks for when the `notificationtype` is disabled
7. Added check for when the channel does not support attachments
8. Added check for when a user has opted out of a notification type


shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -56,6 +56,7 @@ public class NotificationSender: INotificationSender, ITransientDependency
         private readonly IStoredFileService _storedFileService;
         private readonly IBackgroundJobManager _backgroundJobManager;
         private readonly INotificationManager _notificationManager;
+        private readonly IRepository<UserNotificationPreference, Guid> _userNotificationPreference;
```
IvanIlyichev: Please rename field to _userNotificationPreference**Repository** to make code more readable

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -76,7 +77,8 @@ public NotificationSender(INotificationChannelSender channelSender,
                                    IStoredFileService storedFileService,
                                    INotificationSettings notificationSettings,
                                    IBackgroundJobManager backgroundJobManager,
-                                   INotificationManager notificationManager)
+                                   INotificationManager notificationManager,
+                                   IRepository<UserNotificationPreference, Guid> userNotificationPreference)
```
IvanIlyichev: Please rename argument to userNotificationPreference**Repository** to make code more readable

shesha-core/src/Shesha.Application/Notifications/NotificationSender.cs
```diff
@@ -123,8 +126,25 @@ private async Task<List<EmailAttachment>> GetAttachmentsAsync(NotificationMessag
         /// <returns></returns>
         public async Task SendNotification<TData>(NotificationTypeConfig type, Person fromPerson, Person toPerson, TData data, RefListNotificationPriority priority, List<NotificationAttachmentDto> attachments = null, GenericEntityReference triggeringEntity = null, NotificationChannelConfig channel = null) where TData : NotificationData
         {
+            // Check if the notification type is disabled
+            if(type.Disable) return;
+
+            if (type.CanOptOut)
+            {
+                // Check User Notification Preferences
+                var userPreferences = await _userNotificationPreference.GetAllListAsync(
```
IvanIlyichev: Avoid loading of unneeded data from the DB. If you just need to check existence of preferences use **AnyAsync** directly like this:
```c#
var optedOut = await _userNotificationPreference.GetAll().AnyAsync(x => x.User.Id == toPerson.Id && x.NotificationType.Id == type.Id && x.OptOut);
if (optedOut)
   return;
```

shesha-core/src/Shesha.Core/Migrations/M20250120033100.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Shesha.Migrations
+{
+    [Migration(20250120033100)]
+    public class M20250120033100 : Migration
+    {
+        public override void Up()
+        {
+            //fix the foreign key constraint
+            IfDatabase("SqlServer").Execute.Sql(@"
```
IvanIlyichev: 1. Looks like this renaming doesn't make any sense, I think the **old** suffix was used just as temporary solution to keep the data for short period of time until we are sure that new functionality is stable enough. New DB objects should not contain any suffixes
2. If this change is really required it should support both DBMS types: Sql Server and PostgreSql

shesha-core/src/Shesha.Core/Migrations/M20250120095800.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Threading.Tasks;
+
+namespace Shesha.Migrations
+{
+    [Migration(20250120095800)]
+    public class M20250120095800: Migration
+    {
+        public override void Up()
+        {
+            Alter.Table("Core_NotificationChannelConfigs").AddColumn("Core_SupportsAttachment").AsBoolean().Nullable();
```
IvanIlyichev: 1. **Core_** prefix is not required for the **SupportsAttachment** property
2. SupportsAttachment is mapped as not nullable bool but created in the DB as nullable. Use `.NotNullable().SetExistingRowsTo(...)` here

IvanIlyichev (changes requested):
Hi @Tshepo1103-lab. Please check  inline comments

shesha-core/src/Shesha.Core/Migrations/M20250120033100.cs
```diff
@@ -14,15 +14,67 @@ public class M20250120033100 : Migration
...
+                 ALTER TABLE ""Core_NotificationMessageAttachments"" 
+                 DROP CONSTRAINT ""FK_Core_NotificationMessageAttachments_MessageId_Core_NotificationMessages_Id"";
+
+                 ALTER TABLE ""Core_NotificationMessageAttachments"" 
+                 ADD CONSTRAINT ""FK_Core_NotificationMessageAttachments_MessageId_Core_NewNotificationMessages_Id""
+                 FOREIGN KEY (""MessageId"") REFERENCES ""Core_NotificationMessages""(""Id"");
+             ");
+
+            //  Changed the FK Constraint name on the table SM_EscalationRules
+            IfDatabase("SqlServer").Execute.Sql(@"
```
IvanIlyichev: Looks like `SM_EscalationRules` is a project-level table, it shouldn't be handled on the framework level.

shesha-core/src/Shesha.Core/Migrations/M20250120033100.cs
```diff
@@ -14,15 +14,67 @@ public class M20250120033100 : Migration
...
+
+                 ALTER TABLE ""SM_EscalationRules""
+                 ADD CONSTRAINT ""FK_SM_EscalationRules_NotificationId_Core_NewNotifications_Id""
+                 FOREIGN KEY (""NotificationId"") REFERENCES ""Core_Notifications""(""Id"");
+             ");
+
+            //  Changed the FK Constraint name on the table Core_HumanActivityConfigurations
+            IfDatabase("SqlServer").Execute.Sql(@"
+                 ALTER TABLE Core_HumanActivityConfigurations 
+                 DROP CONSTRAINT FK_Core_HumanActivityConfigurations_NotificationOnPositiveResultId_Core_Notifications_Id
```
IvanIlyichev: FK_Core_HumanActivityConfigurations_NotificationOnPositiveResultId_Core_Notifications_Id is missing in the DB. Make sure that you debug migrations on a correct version of the DB, to do it you can use an empty database from [this folder](https://github.com/shesha-io/shesha-framework/tree/main/shesha-starter/database)

shesha-core/src/Shesha.Core/Migrations/M20250120033100.cs
```diff
@@ -14,15 +14,67 @@ public class M20250120033100 : Migration
...
+
+                 ALTER TABLE Core_HumanActivityConfigurations
+                 ADD CONSTRAINT FK_Core_HumanActivityConfigurations_NotificationOnPositiveResultId_Core_NewNotifications_Id
```
IvanIlyichev: As we discussed earlier DB objects shouldn't contain any `New` or `Old` suffixes. You can delete all `Old` objects and use fluent migration syntax without SQL, in this case all names will be generated by a right way

IvanIlyichev (changes requested):
Hi @Tshepo1103-lab. Please check inline comments. Regarding Core_ prefix: I see all columns in the `Core_NotificationChannelConfigs` contain this prefix, they were initially created wrong. Please rename all columns in the table and add `[Prefix(UsePrefixes = false)]` attribute to the `NotificationChannelConfig` class

IvanIlyichev (approved):

PR #2642, opened 2025-01-30
En/notification framework

Tshepo1103-lab:
- Adding the direct send to a person who is not on the system


shesha-core/src/Shesha.Application/Notifications/Emails/EmailChannelSender.cs
```diff
@@ -170,12 +175,7 @@ private MailMessage BuildMessageWith(string fromAddress, string toAddress, strin
                 IsBodyHtml = true,
             };
 
-            if (string.IsNullOrWhiteSpace(fromAddress) || !StringHelper.IsValidEmail(fromAddress))
-            {
-                throw new ArgumentException("Invalid 'from' email address.");
-            }
-
-            message.From = string.IsNullOrWhiteSpace(fromAddress) ? new MailAddress(smtpSettings.DefaultFromAddress) : new MailAddress(fromAddress);
+            message.From = string.IsNullOrWhiteSpace(fromAddress) || !StringHelper.IsValidEmail(fromAddress) ? new MailAddress(smtpSettings.DefaultFromAddress) : new MailAddress(fromAddress);
```
IvanIlyichev: Non-transparent behaviour. Default from address should be applied only in case when `from` is empty or null. And chosen from address should be validated.

shesha-core/src/Shesha.Application/Notifications/INotificationChannelSender.cs
```diff
@@ -13,7 +13,7 @@ namespace Shesha.Notifications
 {
     public interface INotificationChannelSender
     {
-        string GetRecipientId(Person person);
-        Task<SendStatus> SendAsync(Person fromPerson, Person toPerson, NotificationMessage message, string cc, List<EmailAttachment> attachments = null);
+        string GetRecipientId(Person person, string identifier);
+        Task<SendStatus> SendAsync(Person fromPerson, Person toPerson, string identifier, NotificationMessage message, string cc, List<EmailAttachment> attachments = null);
```
IvanIlyichev: Please review arguments. New `identifier` is on the same level as `toPerson` and when specified it replaces `toPerson`. There are at least two options to make the code more generic and more stable:

1. Use two overloaded versions of SendAsync which accepts only one argument: `toPerson` or `identifier`, in this case modified code dis backward compatible with previous version, you don't break previous usages of the `INotificationChannelSender`. But this approach has limitation: if you'll need to add any new logic (e.g. override `fromPerson` by the same way) it will require additional overload of the method
2. Declare a new interface `IMessageParticipant` (or even split it into `IMessageSender` and `IMessageReceiver`) that returns an address (according to the notification channel). Add an overload method like this:
```csharp
Task<SendStatus> SendAsync(IMessageSender from, IMessageReceiver to, string identifier, NotificationMessage message /* other arguments... */);
```
In this case you'll be able to add any implementation without breaking changes and new overloaded methods. First two implementation of interface(s) will be for Person and raw addresses (string). Also these interfaces can be extended to support some sender/receiver related settings (e.g. `DontUseDefaultEmailFromAddress` or `MarkAllMessagesAsUrgent`)

IvanIlyichev (changes requested):
Hi @Tshepo1103-lab. Please check inline comments

PR #2656, opened 2025-02-04
James/en/2207e

James-Baloyi:


shesha-reactjs/src/components/reactTable/tableRow.tsx
```diff
@@ -114,7 +129,26 @@ export const TableRow: FC<ISortableRowProps> = (props) => {
...
+          return (
+            <Fragment>
+              <RowCell onMouseOver={(props) => {
+                setPopupShowing(props);
+              }} getCellRef={(ref) => {
+                handleCellRef(ref.current);
+              }} onMouseLeave={() => {
+                setPopupShowing(false);
+              }} cell={cell} key={cellIndex} row={row.cells} rowIndex={index} />
+              {popupShowing && Boolean(cell.value) && formMode !== 'designer' && showOverflow === true && (innerCellRef.innerText === cell.value?._displayName || innerCellRef.innerText === cell?.value) &&
```
IvanIlyichev, line 141: Cell text can't be used for cell identification purposes. If the table contain two cells with the same rows it'll lead to duplication of popups and these duplicates will not be visible at the same time just because of the same position.

IvanIlyichev: Don't use null propagation where it's not required. Value of `cell.value` already checked before, usage of `cell?.value` or `cell.value?.` is incorrect because `cell` and `cell.value` can't be null/undefined. Code like this may lead to bugs because of bad code readability.

shesha-reactjs/src/components/reactTable/tableRow.tsx
```diff
@@ -64,11 +65,17 @@ export const TableRow: FC<ISortableRowProps> = (props) => {
...
     inlineDisplayComponents,
+    showOverflow,
   } = props;
 
   const { styles } = useStyles();
   const { dragState, setDragState } = useDataTableStore();
   const tableRef = useRef(null);
+  const [popupShowing, setPopupShowing] = useState<Boolean>(false);
+  const [popupPosition, setPopupPosition] = useState<{ x: number; y: number }>({ x: 0, y: 0 });
+  const [innerCellRef, setInnerCellRef] = useState(null);
```
IvanIlyichev, line 76: Proposed change is overloaded by usage of unneeded states, it leads to a ton of re-renders.

IvanIlyichev (changes requested):
Hi @James-Baloyi.
Please check inline comments. 
General comment: rendering of popup should be a part of cell but html element should be outside the table, it allows to overlap table and prevent scrolling and blinking.
You are trying to make change much more complex than it should be.
Try to use [ellipsis example](https://ant.design/components/typography#typography-demo-ellipsis) from AntD as a reference. It can be included into StringCell, see below:
```typescript
export const StringCell = <D extends object, V extends any = any>({ value }: IStringCellProps<D, V>) => {
  const stringValue = typeof(value) === 'string' ? value : value?.toString();
  return (
    <Text
      ellipsis={{ tooltip: value }}
    >
      {stringValue}
    </Text>
  );
};
If there are other type of cells which should support the same functionality (e.g. Entity Reference) then this code can be extracted to a new component, display text (`_displayText` or whatever) from cell should be passed to this new common component. This approach prevents duplication of code and make it more stable
```

PR #2687, opened 2025-02-11
retain the selected component (if any) when previewing a form configu…

teboho:
…ration and coming back to the designer mode

## Summary by CodeRabbit

- **New Features**
  - Introduced a preview mode in the form designer, allowing users to toggle and view forms in preview state.


IvanIlyichev (commented):
Hi @teboho. Please move this logic to the reducer and remove useEffect from the designerMainArea.

IvanIlyichev (changes requested):
Hi @teboho. Please retest it, after switch to the preview mode and back I got an empty properties panel, see below.
Also the `isPreview` flag is used by the `Preview` button only, the designer doesn't take it into account and uses form mode during render. Designer should use the `isPreview` flag otherwise this flag can be removed. Previous implementation (with useEffect) worked because of multiple unneeded re-renders
image

Issue #2560, opened 2024-12-24
Configuration meta-data should be included in the generated Html


ihouvet:
### Overview
A feedback module in being built which would allow end-users to add comments to Shesha pages views. It should be possible from the selected DOM element to determine which is the closest form component, sub-form (if applicable) and rendered form.

See: https://chatgpt.com/share/e/6762e996-5c14-8010-bf1b-58b2d5f4d916 regarding adding custom attributes to HTML elements.

### Acceptance Criteria
1. For form configs: Add a class `sha-f` to the root element of a configured form to help us easily identify configured forms and add custom data attributes e.g. `data-sha-form-id` (then feedback client tool will have to look up the additional required information); `data-sha-form-is-top` indicating if top level form 
2. Similarly for the form components: add `sha-c` class and `data-sha-c-name`, `data-sha-c-type` attributes

teboho:
Given that the extra className is interfering with the styling of the component etc, is there actual need to do it like this if we can still pass the data via data-* attributes 
> For form configs: Add a class sha-f to the root element of a configured form to help us easily identify configured forms and add 
> custom data attributes e.g. data-sha-form-id (then feedback client tool will have to look up the additional required information); 
> data-sha-form-is-top indicating if top level form

teboho:
@hlayisanimazuze making changes to add the form name when the form is a subform

teboho:
When the subForm is in a datalist or datatable


![Image](https://github.com/user-attachments/assets/6ec80602-95a5-4ebd-a806-ebe9e1a25de2)

hlayisanimazuze:
Hi @teboho, here's an example: https://linux-functional-adminportal-test.azurewebsites.net/dynamic/TestModule/urlCustom-details?id=792e7691-0a37-4b65-947c-0076315d5878&firstName=Muriel%20Updatedvvggg&lastName=subform--006

As I mentioned, subform using dynamic form selection mode comes out as undefined. Let me know if it is possible to access the name or not, then we can update the item accordingly

![Image](https://github.com/user-attachments/assets/b3c5bf43-4585-4ed8-a6eb-40309e3b9604)

PR #2702, opened 2025-02-13
added classnames and data-* attributes to be used with the feedback tool

teboho:
Implementing the issue https://github.com/shesha-io/shesha-framework/issues/2560


IvanIlyichev (changes requested):
Hi @teboho. Components wrapper breaks drag&drop functionality in the forms designer. Also it may break custom css classes and styles which are used for elements alignment because after this change all components are rendered inside div

teboho:
> Hi @teboho. Components wrapper breaks drag&drop functionality in the forms designer. Also it may break custom css classes and styles which are used for elements alignment because after this change all components are rendered inside div

Thanks for the feedback @IvanIlyichev I will take further

teboho:
@IvanIlyichev I have applied changes in response to 

> Hi @teboho. Components wrapper breaks drag&drop functionality in the forms designer. Also it may break custom css classes and styles which are used for elements alignment because after this change all components are rendered inside div

---

I have since moved the component wrapper to contain component metadata to the FormComponent. Not using the classNames anymore because the info can be deduced from the data-* attrs](https://github.com/shesha-io/shesha-framework/pull/2702/commits/01d4f3db8bbf0d351b919d2525663abbf5315dce)

- the drag&drop breakage is avoided
- no longer uses extra classnames, that info will be determined from the data-* attributes once it is populated in the browser DOM as was mentioned required in https://github.com/shesha-io/shesha-framework/issues/2560

IvanIlyichev (changes requested):
Hi @teboho. Please try to avoid additional divs in markup, use this example as a reference https://stackoverflow.com/a/75875667

teboho:
> additional divs in marku

teboho:
> Hi @teboho. Please try to avoid additional divs in markup, use this example as a reference https://stackoverflow.com/a/75875667

will make changes

IvanIlyichev (approved):

PR #2935, opened 2025-03-13
Implemented Stress Testing OTP API to allow OTP retrieval via email/cell

mpendulontshangase:
When performing stress testing it is often necessary to do automated bulk registrations of users. Clicking of an email confirmation link or capturing of a registration OTP are often part of the user registration process. It is not possible automate this process during stress testing.
To enable stress testing it is therefore necessary to provide a mechanism for the stress testing scripts to retrieve the OTP/Link through an end-point only made available during stress testing.


IvanIlyichev (changes requested):
Hi @mpendulontshangase. It shouldn't be a part of framework. Please move it to a separate project e.g. Shesha.StressTesting and maybe we even should move it to enterprise to prevent security risks

PR #3255, opened 2025-05-15
Update/collapse panels on preview

teboho:


shesha-reactjs/src/providers/formDesigner/index.tsx
```diff
@@ -1,353 +1,367 @@
...
+      componentRef !== state.present.selectedComponentRef)
+      dispatch(setSelectedComponentAction({ id: componentId, componentRef }));
+      componentInitialization.current = true;
+  }, [dispatch]);
+
+  const setPreviousSelectedComponent = useCallback((componentId: string, componentRef?: MutableRefObject<any>) => {
+    dispatch(setPreviousSelectedComponentAction({ id: componentId, componentRef }));
+  }, [dispatch]);
+
+  useEffect(() => {
```
IvanIlyichev, line 234: Hi @teboho. There is no need to create a separate action to maintain previous selection. Current code should be a part of reducer, otherwise it causes 2 unneeded re-renders and makes the logic more complex. You just need to save and restore selection when the reducer handles `SetFormMode` action.

IvanIlyichev (changes requested):

IvanIlyichev (approved):

Issue #3212, opened 2025-05-01
Theme settings adjustments


hlayisanimazuze:
- Upon saving theme changes, the page resets and navigates back to the general list. Is it not possible to keep the old behavior instead? [Jam link](https://jam.dev/c/9fdf276f-2e57-4c84-b3d2-16d1aa6231f5)
- Please fix the theme parameters and their designs. Currently, all components are touching the edges of the panel they’re rendered on. Restore the same padding we had in the previous implementation. See: https://functional-v0-41-adminportal-test.azurewebsites.net/shesha/settings
- Sidebar section: Unable to change the value. Example: Change Dark to Light, it does not select
- When changing colors in the theme, buttons and texts that are supposed to reflect the changes in real time are not updating. Example: if you change a primary color, the button representing primary color should reflect the changes in real time. See attached video: https://jam.dev/c/5b8c5ad1-69c4-47cc-b126-89c3b4eb1e11
Expected Results from older branch: https://jam.dev/c/1306ab79-d7d5-474a-b75b-f1e4fdd97303

The below screenshot is from v41
![Image](https://github.com/user-attachments/assets/69713e8b-2926-4ca8-8011-3d16dc62b27e)

The below screenshots are from the current changes

![Image](https://github.com/user-attachments/assets/89b74319-eea7-4fee-884c-61ca433dddee)

![Image](https://github.com/user-attachments/assets/46309913-bef8-4537-92f9-5deecb5f440f)

teboho:
# Theme Issues Checklist

## Navigation Issues
- [ ] Fix page reset on saving theme changes
  - [ ] Maintain current page position instead of navigating back to general list
  - [ ] Reference: Jam link provided in ticket

## Layout/Design Issues
- [ ] Restore proper padding in theme components
  - [ ] Fix components touching the edges of their panels
  - [ ] Match padding from previous implementation
  - [ ] Reference: https://functional-v0-41-adminportal-test.azurewebsites.net/shesha/settings

## Functionality Issues
- [ ] Fix sidebar section value changing
  - [ ] Debug inability to change from Dark to Light mode
  - [ ] Ensure selected values persist after selection

## Real-time Preview Issues
- [ ] Fix real-time updates for color changes
  - [ ] Ensure buttons reflect color changes immediately
  - [ ] Ensure text elements update with new colors as applied
  - [ ] Reference: https://jam.dev/c/5b8c5ad1-69c4-47cc-b126-89c3b4eb1e11
[...]

PR #3364, opened 2025-06-06
Update/theme

teboho:
image

https://github.com/shesha-io/shesha-framework/issues/3212

## Summary by CodeRabbit

- **New Features**
	- Added an "Apply Globally" switch in theme settings to control whether text styles are applied globally across the app.
	- Text color can now be applied globally based on the new setting, updating the appearance of text components when enabled.

- **Improvements**
	- Enhanced theme color handling to prevent errors if theme data is missing.
	- Updated styles in theme settings for improved layout and usability.
	- Default text color in theme examples now reflects the current theme setting.

- **Bug Fixes**
	- Corrected placeholder text capitalization in the theme settings form.


IvanIlyichev (commented):
Hi @teboho. You added a new switch `Apply Globally` that is not described in the issue you are referring to #3212. This switch is not required for keeping of previous functionality and it's purpose is not transparent. We should avoid any 'Apply globally'/ 'force' etc. and if we want to configure global and `local` color then we should provide two separate color pickers for it, not a combination of single color picker and `Apply Global` switch

teboho:
Hi @IvanIlyichev thanks for the feedback, I made a different branch and PR that follows the advice here

https://github.com/shesha-io/shesha-framework/pull/3374

PR #3391, opened 2025-06-12
reinsiating old behaviour to the theme editor

teboho:
https://github.com/shesha-io/shesha-framework/issues/3212

## Summary by CodeRabbit

- **New Features**
  - Text elements now dynamically adapt their colors based on the selected theme, providing a more cohesive and customizable appearance.
  - Color pickers in theme settings now support clearing the selected color.

- **Refactor**
  - Improved internal logic for handling text color selection and theming for better maintainability and consistency.
  - Enhanced theme management functions to reduce unnecessary re-renders.

- **Style**
  - Text samples and components now visually reflect theme color changes more accurately.

### DIFF
--- a/shesha-reactjs/src/designer-components/text/genericText.tsx
+++ b/shesha-reactjs/src/designer-components/text/genericText.tsx
@@ -51,10 +51,10 @@ export const GenericText: FC<PropsWithChildren<IGenericTextProps>> = ({
     model.strong,
   ]);
 
-  const remainingColors = ((contentType: ContentType) => {
+  const remainingColors: string | undefined = ((contentType: ContentType) => {
     switch (contentType) {
       case 'custom':
-        return style.color;
+        return style?.color;
       case 'secondary':
         return theme?.text?.secondary;
       case '':
@@ -64,7 +64,7 @@ export const GenericText: FC<PropsWithChildren<IGenericTextProps>> = ({
     }
   })(contentType);
 
-  const chosenType = contentType === 'secondary' ? null : (contentType as BaseType);
+  const chosenType: BaseType | undefined = contentType === 'secondary' ? undefined : (contentType as BaseType);
 
   const baseProps: ITypographyProps = {
     code: model?.code,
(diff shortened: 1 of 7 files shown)


shesha-reactjs/src/designer-components/text/genericText.tsx
```diff
@@ -48,6 +51,21 @@ export const GenericText: FC<PropsWithChildren<IGenericTextProps>> = ({
     model.strong,
   ]);
 
+  const remainingColors: string | undefined = ((contentType: ContentType) => {
```
IvanIlyichev: @teboho, please avoid usage of IIFE where it's not required. It makes code less re adable. Also names if constants should be transparent. You named constant as `remainingColors` but it contains a single color selected by ContentType calculated using IIFE. In addition, part of this calculation is duplicated below, I mean this row: `color: contentType === 'custom' ? style.color : remainingColors,`.
It's much better to declare a pure function (as constant) outside the component that will calculate color by content type and use it in the property assignment. Keep the code clear and simpler to avoid rework and to minimize bugs

shesha-reactjs/src/designer-components/text/index.tsx
```diff
@@ -31,6 +33,7 @@ const TextComponent: IToolboxComponent<ITextTypographyProps> = {
     const backgroundStyles = allStyles?.backgroundStyles;
 
     const additionalStyles: CSSProperties = removeUndefinedProps({
+      ...(theme?.text?.default && { color: theme?.text?.default }),
```
IvanIlyichev: In this line you merge string with object. What is the purpose of it?

shesha-reactjs/src/providers/theme/index.tsx
```diff
@@ -77,21 +76,17 @@ const ThemeProvider: FC<PropsWithChildren<ThemeProviderProps>> = ({
   return (
     <UiStateContext.Provider value={state}>
       <UiActionsContext.Provider
-        value={{
+        value={useMemo(() => ({
```
IvanIlyichev: Hooks shouldn't be used in direct assignment of props. Surprisingly it works but it 100% source of bugs and we can't guarantee how it will be transpiled by React. In theory sequence of hooks execution may change between renders

IvanIlyichev (changes requested):
@teboho see comments inline

shesha-reactjs/src/designer-components/text/index.tsx
teboho: Actually this no longer has purpose with the new switch statement in the generictext file, it was a safety guard rail for default but it is no longer neeeded, I shall remove it.

Issue #3386, opened 2025-06-11
Button group crashes when Entity option changes to URL for Dynamic items


yuliaGradova:
**Describe the bug**
If you configure a button group with dynamic items using the Entity option, and then change it to use a URL, the button group crashes, and you have to click Try Again for the component to render.

"https://jam.dev/c/5f26c2cf-c4a9-4b53-a4a7-d0069fa26fd9"

yuliaGradova:
@IvanIlyichev 
1. The information is pulling from the previously configured Dynamic items = URL though the setting was changed to Dynamic items = Entity
https://linux-functional-adminportal-test.azurewebsites.net/shesha/forms-designer?id=42e80675-ee54-452c-8334-dbb5f3278c39
2. When you click the "Add New Item" button, the added item becomes deselected. If you try to reselect any item, it is automatically deselected again. Pay attention to the dashed border line, which indicates that an item is selected.

IvanIlyichev:
Hi @Tshepo1103-lab. I've checked latest issue mentioned by @yuliaGradova. The reason in sharing of the [useTemplates](https://github.com/shesha-io/shesha-framework/blob/7c7a14c1648370038098763eb539e2d89da68062/shesha-reactjs/src/providers/dynamicActions/implementations/dataSourceDynamicMenu/utils.ts#L10) between two types of datasources: entity and url. If I got it right the entity datasource doesn't use custom urls but `useTemplates` takes it is specified. There shouls be separate hooks for different datasources to avoid conflicts, also make sure that custom url is/not required in the entity datasource and fix settings/logic if required.
PS: I've reassigned issue to you because you an author of this functionality and know details of initial requirements.

Tshepo1103-lab:
Hi @yuliaGradova, the latest comment point 1 has been fixed.

PR #3411, opened 2025-06-17
fix: creating separate hooks to avoid data caching

Tshepo1103-lab:
#3386


IvanIlyichev (approved):
Note: it still requires some refactoring. Dependencies should be handled by hooks themselve. We'll do it later

PR #3464, opened 2025-06-27
Feature/3351/permissions cache on client side

ktsapo:


IvanIlyichev (changes requested):
Hi @ktsapo. Caching should be integrated into [authenticator](https://github.com/shesha-io/shesha-framework/blob/2e0f28e50d2aa37290db59431c6fadaed67c6703/shesha-reactjs/src/providers/auth/authenticator.ts#L148). And I think it even doesn't require any storage, Shesha fetches granted permissions during authentication check and stores list of granted permissions in memory.
Also please revert changes in appsettings.json

Issue #3765, opened 2025-08-27
Number component does not allow saving null value


Tshepiso-R:
### Description
The **Number** component does not handle nullable fields correctly.  When a user clears the value in a number field (declared as nullable) and attempts to save, the backend rejects the request with a format error instead of treating the empty value as `null`.

### Steps to Reproduce
1. Navigate to:  
   [Test Module Form](https://linux-functional-adminportal-test.azurewebsites.net/dynamic/TestModule/new-form-member?id=08808737-f8f6-4de3-aad4-0cc61baae89d)
2. Click **Edit**.
3. Remove the value in the **Targeting Flag** field.
4. Open **Network** tab in DevTools.
5. Click **Save**.

### Actual Result
- The request fails with error:  

### Screenshot

<img width="1920" height="869" alt="Image" src="https://github.com/user-attachments/assets/85e5d7b7-87ba-4901-9497-76bcbf081b4c" />

teboho:
The proposed solution is to coerce null to be zero while we investigate for a long-term solution in the backend. @James-Baloyi @Tshepiso-R 

--- 
This was rejected, the solution must take place in the backend

PR #3782, opened 2025-09-03
Coerce/numberField null/undefined values to zero

teboho:
https://github.com/shesha-io/shesha-framework/issues/3765#issue-3359284844

## Summary by CodeRabbit

* **Bug Fixes**
  * Number inputs now coerce null/undefined to a numeric fallback (uses field min when >0, otherwise 0) and emit that initial value on mount to avoid blank/NaN states.
  * Read-only views consistently format and display coerced numeric values.
  * Controlled/uncontrolled number fields normalize input (strings → numbers, NaN handled) so onChange receives a stable numeric value.

* **UX Improvements**
  * More predictable, consistent behavior for number fields across edit and read-only modes, including high-precision inputs.

### DIFF
--- a/shesha-reactjs/src/designer-components/numberField/control.tsx
+++ b/shesha-reactjs/src/designer-components/numberField/control.tsx
@@ -1,5 +1,5 @@
 import { InputNumber, InputNumberProps } from 'antd';
-import React, { FC } from 'react';
+import React, { FC, useEffect } from 'react';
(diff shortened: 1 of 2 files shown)


teboho:
@coderabbitai review this PR

IvanIlyichev:
Hi @teboho. We must NOT replace null values with zeros

Issue #3748, opened 2025-08-25
Datalist component to support options to save with formData not only inline.


Annahh29:
l would like to use a datalist to render an input field multiple times depending on the data coming from the backend. This input field should always be in edit mode (since a user has to input data) but l do not want it expose the inline save buttons. The reason being, l want to save this data once with the form (its either all data gets saved or nothing). 

The datalist should allow 'Save with FormData' as part of Save mode options.

<img width="378" height="93" alt="Image" src="https://github.com/user-attachments/assets/48a3150c-3420-48c2-b27f-1befda25b018" />.

**Note**: it should be fixed on 0.43, then merge to 0.44 and to the main also

zukidlomo:
1. For existing forms the "Show edit icon" property must be set to True by default.

<img width="658" height="249" alt="Image" src="https://github.com/user-attachments/assets/d9fcb8a1-de54-482a-9f52-3adcfd9babc2" />

yuliaGradova:
@HackGenesis Tested on 0.43 - it's ok. Please merge to 0.44 and main also

yuliaGradova:
@HackGenesis checked in 0.44 - no "Show edit icon" property

<img width="1581" height="726" alt="Image" src="https://github.com/user-attachments/assets/3256f8cf-9c21-414f-b6a0-b0ccdee7d92a" />

zukidlomo:
@HackGenesis Tested on 44 - it's ok. On main there is no "Show Edit Icon" property.

PR #3808, opened 2025-09-09
set default switch state

HackGenesis:
https://github.com/shesha-io/shesha-framework/issues/3748

## Summary by CodeRabbit

* **Bug Fixes**
  * Restores visibility of edit icons in DataList components after migrating existing configurations.
  * Prevents cases where edit icons could be hidden or inconsistent following upgrades.

* **Chores**
  * Improves migration robustness to ensure UI settings for DataList are consistently preserved.


HackGenesis:
> Hi @HackGenesis. If I got it right you are trying to set a default value for a new setting. If so, it shouldn't be done as part of configuration form, use migrations instead

@IvanIlyichev will do so thanks 👌🏼

Issue #3648, opened 2025-07-29
Implement File Download Tracking to Control UI Component Behavior


Tshepiso-R:
**Description:**
Introduce a mechanism to monitor and detect whether a user has downloaded a file exposed through the File and FileList components. This download status should be made programmatically available so that it can be used to trigger other actions such as hiding, disabling, or enabling other UI elements conditionally.

**Acceptance Criteria:**

- [ ] A file download event is captured accurately when the user downloads a file from File and FileList.

yuliaGradova:
Additional requirements:
Different styles (colour) should be implemented to identify not downloaded files.

**Note**: the same feature was requested by DHET for 0.41 branch

yuliaGradova:
1. Click on Press to upload and select the file. **Error:** the system checks if it's dowloaded immediately after uploading
The same when user uploads the 2nd file - the system already checks both files

<img width="1106" height="847" alt="Image" src="https://github.com/user-attachments/assets/9000e3ab-a639-4a31-a01a-710e64f3abda" />

2. User1 is dowloaded Zip. Login as user2 and open the record with previously downloaded zip (downloaded by user1). **Error**: for user2 the system identifies zip as already downloaded

<img width="1251" height="913" alt="Image" src="https://github.com/user-attachments/assets/549289c9-a66b-4145-ba58-a9940e12eb5a" />

[My form](https://linux-functional-v0-43-adminportal-test.azurewebsites.net/shesha/forms-designer?id=5d9a4603-c9af-4b7e-8aa2-67db7f342a29)

yuliaGradova:
> 1. Click on Press to upload and select the file. **Error:** the system checks if it's dowloaded immediately after uploading
>    The same when user uploads the 2nd file - the system already checks both files

<img width="1477" height="1005" alt="Image" src="https://github.com/user-attachments/assets/7896c78d-13a9-4268-8f72-cdf11642d9cc" />

**Not fixed**

2. After download files and full page refresh the requests are duplicated

<img width="1386" height="1071" alt="Image" src="https://github.com/user-attachments/assets/80dfdf66-e157-4d37-8b12-f2981c699c33" />

[My form](https://linux-functional-v0-43-adminportal-test.azurewebsites.net/shesha/forms-designer?id=72d5fd5a-bccc-40bb-92ac-bf10934846ce)

yuliaGradova:
Tested on 0.44 and main:
[...]

PR #3818, opened 2025-09-10
Thulasizwe/f/43/3769

czwe-01:
#3648

## Summary by CodeRabbit

- New Features
  - Attachments editor now syncs file lists with parent value and reliably runs “on file changed” scripts with current files.
  - Provider can initialize file lists from existing values, supports single-file download, and marks files as downloaded (per-file or all).
  - Settings: added “Property name” field with autocomplete; code editor gains predefined constants (value, event).

- Bug Fixes
  - More consistent deletion and state updates; onChange emits updates only when real changes occur.
  - Removed legacy file-list callback to prevent duplicate events.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,6 +1,6 @@
 # Release Notes
-💪 Enhancements
-- GenericEntityReferences support for JsonEntity
 
 🐞 Bug Fixes
-- Fixed support of custom column names in SoftDelete filter for one-to-may relations
+- Number component does not allow saving null value
+- Fix export forms problem
+- Fix job's deactivation
--- a/shesha-core/src/Shesha.Application/ConfigurationItems/ConfigurationItemAppService.cs
+++ b/shesha-core/src/Shesha.Application/ConfigurationItems/ConfigurationItemAppService.cs
@@ -4,6 +4,7 @@ using Abp.Domain.Entities;
 using Abp.Domain.Repositories;
 using Abp.Runtime.Validation;
 using Microsoft.AspNetCore.Mvc;
+using Microsoft.Extensions.Options;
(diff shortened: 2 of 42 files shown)


czwe-01:
> Hi @czwe-01
> 
> 1. Usages of `value` and `state.fileList` are mixed up
> 2. Duplicated code
> 3. `isDownloadedByCurrentUser` flag should be fetched with the file list at once
> 4. Attached issue is not related to changes implemented in this PR
> 5. PR changes behaviour of other components: text field, text area and chevron. Please change them as part of corresponding PR and issue
> 6. Purpose of the `property` setting as par tof attachments editor is not clear and may break existing functionality (validation). I'm not an author of this validation and it looks strange but your change shouldn't break this functionality

Hi @IvanIlyichev, thank you for the review. 
For the last comment about the propertyName
We have implemented updates to the propertyName setting to enhance functionality for users who need to utilize the fileList property within their scripts.
Implementation Details:
When propertyName is specified:
The fileList property becomes accessible within the data object in code editors.
This enables direct manipulation and reference of file data within custom scripts.
The property is explicitly named and available for programmatic use.
When propertyName is left blank:
The system utilizes an internal ghost propertyName.
This approach prevents potential submission errors that could occur when including the property in form data.
Form submissions can proceed without interference from the file property.
The previous implementation likely excluded this property from the data object to avoid submission conflicts and potential errors during form processing. The updated approach provides flexibility by allowing developers to choose between explicit property access (when a name is provided) and seamless form submission (when left blank).
This enhancement maintains backward compatibility while expanding the component's capabilities for advanced use cases.

IvanIlyichev (approved):

PR #3831, opened 2025-09-12
Tshepo/en/download tracking

Tshepo1103-lab:
#3648

## Summary by CodeRabbit

* **New Features**
  * File lists and file details now show whether you (the current user) have previously downloaded each file; downloads are tracked per user.

* **Improvements**
  * Thumbnail delivery rewritten with better caching, 304 handling and more reliable generation.
  * ZIP downloads now ensure download status is recorded for included files.
  * Base64 file endpoint now returns a JSON object with the encoded string.

* **Bug Fixes**
  * Minor fix for an edge case affecting file deletion.

### DIFF
--- a/shesha-core/src/Shesha.Application/StoredFiles/StoredFileController.cs
+++ b/shesha-core/src/Shesha.Application/StoredFiles/StoredFileController.cs
@@ -7,6 +7,7 @@ using Abp.Reflection;
 using Abp.Runtime.Session;
 using Abp.Runtime.Validation;
 using Abp.UI;
+using DocumentFormat.OpenXml.Spreadsheet;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.ModelBinding;
 using Shesha.Authorization;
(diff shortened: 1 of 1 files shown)


shesha-core/src/Shesha.Application/StoredFiles/StoredFileController.cs
```diff
@@ -532,35 +532,51 @@ public async Task<FileStreamResult> DownloadZipAsync([FromQuery] FilesListInput
...
+
+            var fileVersions = input.FilesCategory.IsNullOrEmpty()
+                ? await _fileService.GetLastVersionsOfAttachmentsAsync(id, type)
+                : await _fileService.GetLastVersionsOfAttachmentsAsync(id, type, input.FilesCategory.ToCamelCase());
+
+            var currentUserId = _abpSession.UserId;
+            if (currentUserId == null)
+                return fileVersions.Select(v => GetFileDto(v)).ToList();
+
+            var downloadedFileIds = await _fileVersionDownloadRepository.GetAll()
```
IvanIlyichev: @Tshepo1103-lab please avoid excessive data fetching. This row fetches identifiers of all files downloaded by the user

Tshepo1103-lab: @IvanIlyichev I have added the condition to make sure that I am only fetching files that are part of the fileversions, it only checks against the latest versions of the files.

shesha-core/src/Shesha.Application/StoredFiles/StoredFileController.cs
```diff
@@ -63,42 +63,42 @@ TypeFinder typeFinder
...
         }
 
         [HttpGet, Route("Download")]
         public async Task<ActionResult> DownloadAsync(Guid id, int? versionNo)
         {
             var fileVersion = await GetStoredFileVersionAsync(id, versionNo);
 
-            if (fileVersion.Id.ToString().ToLower() == HttpContext.Request.Headers.IfNoneMatch.ToString().ToLower())
-            {
-                await _fileService.MarkDownloadedAsync(fileVersion);
```
IvanIlyichev, line 78: `MarkDownloadedAsync` is called only on a second download attempt. It should be moved outside the condition

IvanIlyichev (changes requested):

shesha-core/src/Shesha.Application/StoredFiles/StoredFileController.cs
Tshepo1103-lab: @Tshepo1103-lab I have the `MarkDownloadedAsync` on line 86 as well. Initially, I only had it outside, but I realized that if I repeat the process, it won’t be marked as downloaded, that is when I added on line 78

IvanIlyichev: @Tshepo1103-lab It's better to make a single call before line 76.

PR #3841, opened 2025-09-17
Thulasizwe/f/43/3769

czwe-01:
#3648
https://github.com/shesha-io/shesha-framework/issues/3648#issuecomment-3298274189

## Summary by CodeRabbit

* **New Features**
  * Added an "On Download" code-editor action in the attachments editor to run scripts when files are downloaded.
  * Exposed an onDownload callback in the attachments component and stored-files provider so download events surface to scripts with form/page context.
  * Added a centralized script-execution wrapper to ensure scripts receive consistent context for file events.

* **Bug Fixes**
  * onChange now fires for meaningful file-list updates; download-related state changes use onDownload.

* **Refactor**
  * Centralized file-list updates with new helper utilities for consistent upload/delete/download state handling.


IvanIlyichev (changes requested):
Hi @czwe-01. Event handlers like `onChange` and `onDownload` must only be called in direct response to a user action or a clear event, not from within React useEffect hooks. Using effects for this purpose makes the component's behavior unpredictable and difficult to control, particularly when using refs (like `shouldTriggerOnDownloadRef`). The current implementation relies on anti-patterns and needs a significant redesign.

shesha-reactjs/src/providers/storedFiles/index.tsx
```diff
@@ -218,9 +215,17 @@ const StoredFilesProvider: FC<PropsWithChildren<IStoredFilesProviderProps>> = ({
...
+        const nextList = (state.fileList ?? []).filter(({ id, uid }) => id !== fileIdToDelete && uid !== fileIdToDelete);
+        onChange?.(nextList);
+        if (typeof addDelayedUpdate === 'function') {
+          removeDelayedUpdate(STORED_FILES_DELAYED_UPDATE, fileIdToDelete);
+        }; 
       })
-      .catch(() => deleteFileError(fileIdToDelete));
+      .catch((e) => {
+        deleteFileError(fileIdToDelete);
+        message.error('Failed to delete file');
```
IvanIlyichev: Visual parts should not be added to the provider. All messages/notifications should be implemented one level up (in a renderer). I see there is another `message` call which was added earlier and it's incorrect too, we'll remove it later I hope. For not let's not create more issues for further refactoring. If you want to handle error, add new event and show message in a handler

shesha-reactjs/src/providers/storedFiles/index.tsx
```diff
@@ -181,13 +169,19 @@ const StoredFilesProvider: FC<PropsWithChildren<IStoredFilesProviderProps>> = ({
...
     dispatch(uploadFileRequestAction(newFile));
 
     uploadFileHttp(uploadFileEndpoint, formData)
       .then((response) => {
         const responseFile = response.result as IStoredFile;
         responseFile.uid = newFile.uid;
         dispatch(uploadFileSuccessAction({ ...responseFile }));
+        // Compute next list after success (replace by uid and set uid=id as reducer does)
+        const updatedFile = fileReducer(responseFile);
+        dispatch(uploadFileSuccessAction({ ...updatedFile }));
```
IvanIlyichev: Duplicate call of the same action with no reasons.

shesha-reactjs/src/providers/storedFiles/index.tsx
```diff
@@ -218,9 +215,17 @@ const StoredFilesProvider: FC<PropsWithChildren<IStoredFilesProviderProps>> = ({
     deleteFileHttp({ id: fileIdToDelete })
       .then(() => {
         deleteFileSuccess(fileIdToDelete);
-        if (typeof addDelayedUpdate === 'function') removeDelayedUpdate(STORED_FILES_DELAYED_UPDATE, fileIdToDelete);
+        const nextList = (state.fileList ?? []).filter(({ id, uid }) => id !== fileIdToDelete && uid !== fileIdToDelete);
```
IvanIlyichev: General comment for all calculations like this. If I you need to call some shared logic that manupilates a list of files - make a functions and reuse them in both reducer and provider. Current implementation contains a lot of duplicates in both provider and reducer.

shesha-reactjs/src/designer-components/attachmentsEditor/attachmentsEditor.tsx
```diff
@@ -62,6 +63,23 @@ const AttachmentsEditor: IToolboxComponent<IAttachmentsEditorProps> = {
 
     const enabled = !model.readOnly;
 
+    const executeScript = (script, value, onChange?) => {
```
IvanIlyichev: There are no calls which pass the `onChange` argument.  Please avoid dead code, it makes code less readable and complicates its support

IvanIlyichev (changes requested):

IvanIlyichev (approved):

PR #3916, opened 2025-09-30
James/en/3889

James-Baloyi:


shesha-reactjs/src/components/dataTable/index.tsx
```diff
@@ -669,14 +672,100 @@ export const DataTable: FC<Partial<IIndexTableProps>> = ({
...
+            argumentsEvaluationContext: evaluationContext,
+          });
+        } catch (error) {
+          console.error('OnAfterRowReorder event error:', error);
+          // Note: We don't throw here as the reorder has already completed successfully
+        }
+      }
+    } catch (error) {
+      // Reset to original order on API error
+      payload.applyOrder(oldData);
```
IvanIlyichev: @James-Baloyi. Rollback action `payload.applyOrder` already executed in the respository level. Additional rollback here unneeded (case 1 below) or may lead to inconsistency of the table state (case 2).
case 1: repository failed to applied new order. In this case `payload.applyOrder` will be executed twice.
case 2: repository successfull applied new order but execution of `onAfterRowReorder` event failed. In this case new order is persisted to the DB but table shows previous order

IvanIlyichev (changes requested):

IvanIlyichev (approved):

Issue #4073, opened 2025-10-27
Expose `dynamicItem`  variable on Profile Dropdown


Tshepiso-R:
## **Description**
When you use item button and you console dynamicItem on your configuration,upon clicking the button,  the error " ReferenceError: dynamicItem is not defined" is logged. This only occures on the profile dropdown. When the very same configuration is used on a normal button, the logs are sgowing successfully
 
---

## **Steps to Reproduce**
1. Open to the application header configuration form.    
3. Click the **Profile Dropdown**, then select **Configure Button Group**.  
4. Click **Add New Group**, then **Add Item**.  
5. Under **Item Type**, select **Dynamic Item(s)**, then choose **Entity**.  
6. For **Entity Type**, select **Organisation (shesha.domain.organisation)**.  
7. For **Label Property**, select **name**.  
8. For **Button Type**, select **Default**.  
9. Under **Action Configuration**, select **Common > Execute Script**.  
10. In the script editor, enter the following code:  
    ```javascript
    console.log("dynamic item: ", dynamicItem);
    ```
11. Save the configuration.  
12. Refresh the page.  
13. Click any one of the organisations displayed under the dynamic items in the header.

---

## **Expected Behavior**
When a user clicks on an organisation under the dynamic menu items, the console should log the selected organisation’s details, for example:
```javascript
dynamic item: { id: "123", name: "ABC Organisation", ... }
```
## **Actual Behavior**
Instead of logging the organisation details, an error appears in the browser console.  
The logged error message indicates that `dynamicItem` is either undefined or inaccessible within the script execution context.

### **Error Example**
Uncaught ReferenceError: dynamicItem is not defined
at <anonymous>:1:27

**Documentation Reference:**  
[https://docs.shesha.io/docs/front-end-basics/how-to-guides/dynamic-menu-items](https://docs.shesha.io/docs/front-end-basics/how-to-guides/dynamic-menu-items)

## Merge Notes
Please merge to 43, 44 and main

Tshepiso-R:
Hi @HackGenesis, please merge on main and 44

Tshepiso-R:
@HackGenesis please address the crashing bug linked to this issue first.
- https://github.com/shesha-io/shesha-framework/issues/4123

yuliaGradova:
The issue is actual again.

1. Open to the application header configuration form.
2. Click the Profile Dropdown, then select Configure Button Group.
3. Click Add New Group, then Add Item.
[...]

PR #4122, opened 2025-11-03
expose dynamic items

HackGenesis:
https://github.com/shesha-io/shesha-framework/issues/4073

## Summary by CodeRabbit

* **Refactor**
  * Profile dropdown actions now receive richer runtime context, so menu actions can evaluate and execute using additional dynamic item data.
  * Configurable buttons and button item definitions have stricter typing for the dynamic item payload, improving reliability when actions depend on item details.

### DIFF
--- a/shesha-reactjs/src/designer-components/profileDropdown/utils.tsx
+++ b/shesha-reactjs/src/designer-components/profileDropdown/utils.tsx
@@ -1,5 +1,4 @@
 import {
-  IButtonItem,
   IConfigurableActionConfiguration,
   IconType,
   IHeaderAction,
@@ -10,6 +9,7 @@ import {
   ButtonGroupItemProps,
   IButtonGroup,
   isGroup,
+  isButtonItem,
 } from '@/providers/buttonGroupConfigurator/models';
 import { IFullAuditedEntity } from '@/publicJsApis/entities';
 import { IAuthenticator } from '@/providers/auth';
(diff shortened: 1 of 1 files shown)


shesha-reactjs/src/designer-components/profileDropdown/utils.tsx
```diff
@@ -55,6 +56,7 @@ export const getMenuItem = (
   return visibleItems.map((item) => {
     const { id, icon, label } = item;
     const childItems = isGroup(item) ? (item as IButtonGroup).childItems : undefined;
+    const dynamicItem = (item as IButtonItem)?.dynamicItem;
```
IvanIlyichev: @HackGenesis please use type guard instead of cast. For this case there is an existing guard `isButtonItem`

shesha-reactjs/src/designer-components/profileDropdown/utils.tsx
```diff
@@ -64,7 +66,7 @@ export const getMenuItem = (
         </Fragment>
       ),
       children: childItems ? getMenuItem(childItems, execute, visibilityChecker) : undefined,
-      onClick: () => execute((item as IButtonItem)?.actionConfiguration),
+      onClick: () => execute((item as IButtonItem)?.actionConfiguration, dynamicItem),
```
IvanIlyichev: Use type guard and don't call execute for groups

IvanIlyichev (approved):

Issue #4049, opened 2025-10-22
File with uppercase extension not downloadable


Tshepiso-R:
**Description:**  
When a file is uploaded with an uppercase extension (e.g., FILE.CSV instead of file.csv), the attachment is not downloadable from the file list. The system seems to be case-sensitive when handling file types, which causes the download process to fail.

**Steps to Reproduce**

1. Navigate to the [form](https://linux-functional-v0-43-adminportal-test.azurewebsites.net/dynamic/Boxfusion.SheshaFunctionalTests.Common/DetailsMemberFilelIst?id=c2f8b8a4-9e9c-4575-ae0c-24fc83fcf32d)
2. Upload a file with an uppercase extension (e.g., REPORT.CSV or DATA.PDF).
3. Open your browser network
4. Attempt to download the uploaded file from the file list.

**Expected Result:**  
The file should download successfully.

**Actual Result:**  
The file is not downloadable.

File to test with:

[Interface Report 22 October 2025 (3).CSV](https://github.com/user-attachments/files/23056323/Interface.Report.22.October.2025.3.CSV)

Tshepiso-R:
@Tshepo1103-lab  @czwe-01 Please merge to 0.44 and main

yuliaGradova:
Fixed on 0.43 branch. Need to be merged to 0.44 and main

PR #4493, opened 2026-02-05
Normalize file type extensions to lowercase in StoredFileController

Tshepo1103-lab:
#4049 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file path resolution handling for cloud storage to properly respect case sensitivity when locating files.


IvanIlyichev (changes requested):
Hi @Tshepo1103-lab. Proposed changes don't fix issue #4049.
The real reason of that issue in handling of blob name in the AzureStoredFileService
We shouln't change names of uploaded files, the user should be able to upload and download files as is without any names conversion

PR #4510, opened 2026-02-11
Fix page size change handling in TablePaging component to use current…

Tshepo1103-lab:
… page instead of defaulting to 1.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents unexpected page reset on initial load when the page size selector is shown/hidden.
  * Pagination no longer auto-triggers a page size change on first render, improving stability and predictability.

### DIFF
--- a/shesha-reactjs/src/components/tablePager/tablePaging.tsx
+++ b/shesha-reactjs/src/components/tablePager/tablePaging.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useEffect } from 'react';
+import React, { FC } from 'react';
 import { Pagination } from 'antd';
 import { useMedia } from 'react-use';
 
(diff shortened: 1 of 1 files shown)


Tshepo1103-lab:
#4511

shesha-reactjs/src/components/tablePager/tablePaging.tsx
```diff
@@ -63,7 +63,7 @@ export const TablePaging: FC<ITablePagerBaseProps> = ({
   };
 
   useEffect(() => {
-    if (!isNaN(selectedPageSize)) onShowSizeChange(1, selectedPageSize);
+    if (!isNaN(selectedPageSize)) onShowSizeChange(currentPage, selectedPageSize);
```
IvanIlyichev: Hi @Tshepo1103-lab . Pager should not silently change table state without user action. Initial usage of `useEffect` in here is incorrect. Try to remove `useEffect` and fix it on the datatable level, I believe it already restores the state

PR #4671, opened 2026-03-20
Hotfix/4624/swagger UI exposed in production no permission check

ktsapo:
## Summary by CodeRabbit

* **New Features**
  * Add a security setting to enable/disable Swagger UI (enabled by default).
  * Middleware enforces Swagger UI access, returning 403 when disabled.
  * API home endpoint now conditionally redirects to Swagger or returns "API is running" based on that setting.

* **Bug Fixes**
  * Improved API schema ID generation to avoid collisions for certain system types.

### DIFF
--- a/shesha-core/src/Shesha.Framework/Swagger/SwaggerUiAccessMiddleware.cs
+++ b/shesha-core/src/Shesha.Framework/Swagger/SwaggerUiAccessMiddleware.cs
@@ -18,9 +18,7 @@ namespace Shesha.Swagger
         public async Task InvokeAsync(HttpContext context, RequestDelegate next)
         {
             var path = context.Request.Path;
-            var isSwaggerUi = path.StartsWithSegments("/swagger", StringComparison.OrdinalIgnoreCase)
-                              && path.Value != null
-                              && !path.Value.EndsWith(".json", StringComparison.OrdinalIgnoreCase);
+            var isSwaggerUi = path.Value != null && path.StartsWithSegments("/swagger", StringComparison.OrdinalIgnoreCase);
 
             if (isSwaggerUi)
             {


shesha-core/src/Shesha.Framework/Swagger/SwaggerUiAccessMiddleware.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            var path = context.Request.Path;
+            var isSwaggerUi = path.StartsWithSegments("/swagger", StringComparison.OrdinalIgnoreCase)
+                && !path.Value.EndsWith(".json", StringComparison.OrdinalIgnoreCase);
+
+            if (isSwaggerUi)
+            {
+                if (context.RequestServices.GetService(typeof(ISecuritySettings)) is ISecuritySettings securitySettings)
+                {
+                    var settings = await securitySettings.SecuritySettings.GetValueAsync();
```
IvanIlyichev, line 33: @ktsapo Swagger should be disabled when settings are not available due to any reasons

shesha-core/src/Shesha.Framework/Swagger/SwaggerUiAccessMiddleware.cs
```diff
@@ -0,0 +1,53 @@
...
+    public class SwaggerUiAccessMiddleware: IMiddleware, ITransientDependency
+    {
+        public ILogger Logger { get; set; } = NullLogger.Instance;
+
+        public async Task InvokeAsync(HttpContext context, RequestDelegate next)
+        {
+            var path = context.Request.Path;
+            var isSwaggerUi = path.StartsWithSegments("/swagger", StringComparison.OrdinalIgnoreCase)
+                              && path.Value != null
+                              && !path.Value.EndsWith(".json", StringComparison.OrdinalIgnoreCase);
```
IvanIlyichev: @ktsapo. All urls which starts with `/swagger` should be disabled otherwise it's possiblke to get access to API schema using direct links like `/swagger/service:Area/swagger.json`

Issue #4341, opened 2025-12-10
Unable to Configure Models on Non-Template Forms - Dialog Buttons Not Working


Tshepiso-R:
## Description
When creating a form without using a template, the action buttons (OK/Save) in the Settings dialog become non-responsive after attempting to configure a model type. This issue is specific to the non-template form creation workflow and does not affect forms created using templates.

**Root Cause**: The **Appearance tab** has a mandatory **Layout** field that is not being validated or communicated to the user. The dialog silently prevents submission due to this missing required field without providing any validation feedback or error messages.

## Steps to Reproduce
URL: [Form created without selecting a template](https://linux-functional-v0-44-adminportal-test.azurewebsites.net/shesha/forms-designer?id=dc6ea4c6-4d0c-4401-96b2-544d2a3687c2)

1. Click on **Settings** button
2. Search for any entity in the model type selector
3. Select and configure the model type
4. Attempt to click **OK** (or equivalent action button)

## Actual Results
- Action buttons remain completely non-responsive
- User is not informed that the Layout field is required
- No visual indication of which field is blocking submission

## Expected Results
**Either:**
- **Default Layout should be set automatically** when creating forms without templates (matches template behavior), OR
- **User should be told to configure Layout** with clear validation errors, field highlighting, and navigation to Appearance tab

James-Baloyi:
This is a .43 issue. Check if sorted

James-Baloyi:
PR Merged. Waiting for build before pushing to testing

PR #4728, opened 2026-04-02
default layout for form settings modal

James-Baloyi:
#4341


IvanIlyichev (changes requested):
Hi @James-Baloyi. Default value should not be set on component level, it's part of legacy functionality that will be removed. This change should be applied on the `formSettingsMigrations`

James-Baloyi:
@IvanIlyichev updated the PR

IvanIlyichev (approved):

Issue #5162, opened 2026-07-15
Creating form from template doesn't work - empty table with templates


yuliaGradova:
When attempting to create a form from a template, the templates table is completely empty. `api/services/app/Entities/GetAll` fails with error `GetAllAsync is not implemented for entity of type Shesha.Domain.FormConfiguration`

**To Reproduce**
1. Navigate to Create Form
2. Select 'Use Template'
3. Observe the templates table is empty
4. Open the browser console or network tab
5. Observe the error: `GetAllAsync is not implemented for entity of type Shesha.Domain.FormConfiguration`

**Expected behavior**
The templates table should display all available form templates

**Screenshots**

<img width="2077" height="964" alt="Image" src="https://github.com/user-attachments/assets/7a290276-8563-43f8-8196-84ce59cbcc5d" />

**Additional Notes**
The same problem when try to select Form in Entity Reference. So it means the problem with API

<img width="2559" height="809" alt="Image" src="https://github.com/user-attachments/assets/726746d8-7bc0-4330-b409-05d056897924" />

Kulani-MM:
Reopening. Selecting a template now works, but proceeding to the **Configurations** step fails:

Error: `Unknown configuration item identifier inside getCachedConfigAsync / getFormAsync`

<img width="2996" height="1648" alt="Image" src="https://github.com/user-attachments/assets/fe508863-b7ba-41bd-968f-aa5412f767c8" />


The same problem with entity reference. 
- Open [the form](https://linux-functional-adminportal-test.azurewebsites.net/dynamic/Boxfusion.SheshaFunctionalTests.ModuleA/PersonDetailsView24?id=80aa3e9a-0fff-4551-ab30-0c465e93db69)
- When I click on Address on details view the form is empty and `Unknown configuration item identifier` error though selected form existed

<img width="2063" height="1195" alt="Image" src="https://github.com/user-attachments/assets/ffc17d03-f1c1-425c-8ebd-692e7d26227b" />

yuliaGradova:
@

PR #5192, opened 2026-07-23
James/b/5162

James-Baloyi:
#5162 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved configuration item identifier handling, including whitespace-only values and normalization for name/module and nested inputs.
  - Enhanced configuration lookup behavior with better cache/request de-duplication and clearer “not found” errors.
  - Fixed form template loading for both GUIDs and module/name references.
  - Improved autocomplete selection syncing and prevented focus loss in multi-select mode.
  - Safer handling of non-configurable data-bound properties and early handling of empty entity types.
- **Configuration**
  - Added a configuration migration containing the form creation template.
- **Performance**
  - Improved URL repository memoization stability.


shesha-reactjs/src/interfaces/configurableItems.ts
```diff
@@ -1,25 +1,46 @@
...
   return isConfigurableItemFullName(value)
-    ? (value.module === null ? value.name : `${value.module}:${value.name}`)
+    ? (isNullOrWhiteSpace(value.module) ? value.name : `${value.module}:${value.name}`)
     : value;
 };
+
+const canonicalModule = (module: unknown): string | null =>
+  typeof module === "string" && !isNullOrWhiteSpace(module) ? module : null;
+
+export const normalizeConfigurableItemIdentifier = (value: unknown): ConfigurableItemIdentifier | undefined => {
```
IvanIlyichev: The normalizeConfigurableItemIdentifier function handles a data type that is not declared in the project. If this type genuinely needs to be handled, it must be explicitly declared, and the corresponding calls must also be present. Otherwise, the code is both pointless and fragile, and it will be removed or modified during any subsequent revision.

shesha-reactjs/src/providers/configurationItemsLoader/configurationLoader.ts
```diff
@@ -313,7 +313,7 @@ export class ConfigurationLoader implements IConfigurationLoader {
       return await this.getCachedConfigAsync<TConfigDto>({ ...args, id: lookup });
     }
 
-    throw new Error('Unknown configuration item identifier');
+    throw new Error(`Unknown configuration item identifier: ${JSON.stringify(id)}`);
```
IvanIlyichev: Within try-catch blocks, it is recommended to either use a safe JSON serialization approach or, alternatively, pass the data directly as an error parameter without any serialization.

shesha-reactjs/src/providers/sheshaApplication/publicApi/forms/api.ts
```diff
@@ -46,15 +46,29 @@ export class FormsApi implements IFormsApi {
     this._entityMetadataHelper = new FormMetadataHelper(metadataDispatcher);
   }
 
+  // expects an already-trimmed, non-empty identifier; returns undefined when it
+  // resolves to an empty name (so an empty GetByName request is never issued)
+  private buildTemplateUrl = (trimmed: string): string | undefined => {
+    const isGuid = /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/.test(trimmed);
```
IvanIlyichev: Guid string validation is currently duplicated. If this check is genuinely necessary, the existing `isValidGuid()` validation function should be used instead.

shesha-reactjs/src/interfaces/configurableItems.ts
```diff
@@ -1,25 +1,46 @@
...
-  readonly module: string | null;
+  readonly module?: string | null;
 }
 
 export type ConfigurableItemUid = string;
 export type ConfigurableItemIdentifier = ConfigurableItemFullName | ConfigurableItemUid;
 
-export const isConfigurableItemRawId = (formId: ConfigurableItemIdentifier): formId is ConfigurableItemUid => {
-  return isDefined(formId) && typeof formId === 'string';
+export const isConfigurableItemRawId = (formId: unknown): formId is ConfigurableItemUid => {
```
IvanIlyichev: Changing the argument type to unknown in the isConfigurableItemRawId type guard will make it prone to misuse, as some checks will be effectively disabled. The same applies to isConfigurableItemFullName(), which I will update myself later.

shesha-reactjs/src/providers/sheshaApplication/publicApi/forms/api.ts
```diff
@@ -46,15 +46,29 @@ export class FormsApi implements IFormsApi {
     this._entityMetadataHelper = new FormMetadataHelper(metadataDispatcher);
   }
 
+  // expects an already-trimmed, non-empty identifier; returns undefined when it
+  // resolves to an empty name (so an empty GetByName request is never issued)
+  private buildTemplateUrl = (trimmed: string): string | undefined => {
```
IvanIlyichev: The argument name trimmed in buildTemplateUrl is inappropriate. Argument names should clearly reflect their semantic purpose rather than their formatting.

shesha-reactjs/src/providers/sheshaApplication/publicApi/forms/api.ts
```diff
@@ -46,15 +46,29 @@ export class FormsApi implements IFormsApi {
...
   }
 
+  // expects an already-trimmed, non-empty identifier; returns undefined when it
+  // resolves to an empty name (so an empty GetByName request is never issued)
+  private buildTemplateUrl = (trimmed: string): string | undefined => {
+    const isGuid = /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/.test(trimmed);
+    if (isGuid)
+      return `/api/services/Shesha/FormConfiguration/Get?${qs.stringify({ id: trimmed })}`;
+
+    const separatorIndex = trimmed.search(/[/:]/);
```
IvanIlyichev: Manual parsing of the identifier is pointless and undocumented (both in the interface description and in the public API). If this functionality is truly required, the proper approach would be to add support for two data types - string and object - and update the documentation accordingly.

shesha-reactjs/src/interfaces/configurableItems.ts
```diff
@@ -1,25 +1,46 @@
-import { isDefined } from "@/utils/nullables";
+import { isDefined, isNullOrWhiteSpace } from "@/utils/nullables";
 
 export interface ConfigurableItemFullName {
   readonly name: string;
-  readonly module: string | null;
+  readonly module?: string | null;
```
IvanIlyichev: Module is a mandatory property. It must be either a string or null (the latter is only for legacy forms, which are not even available in the current version). We should not relax this type constraint.

IvanIlyichev (changes requested):

IvanIlyichev (changes requested):
Hi @James-Baloyi. Please clarify the purpose of `PersistedConfigurableItemIdentifier`, `PersistedConfigurableItemRawId`, and `PersistedConfigurableItemFullName`. Currently, these are used only within IConfigurationLoader.getCurrentConfigAsync() and have no other references in code or form configurations. If they are indeed unused, removing them would reduce unnecessary complexity. Additionally, please mark any previous comments as resolved if they have been addressed.

James-Baloyi:
@IvanIlyichev 

They were added in commit 38978b40a in response to your earlier review comment ("the function handles a data type that is not declared in the project… it must be explicitly declared"). Before that, normalizeConfigurableItemIdentifier took unknown and did the same shape sniffing untyped.

"The normalizeConfigurableItemIdentifier function handles a data type that is not declared in the project. If this type genuinely needs to be handled, it must be explicitly declared, and the corresponding calls must also be present. Otherwise, the code is both pointless and fragile…"

IvanIlyichev:
> @IvanIlyichev
> 
> They were added in commit [38978b4](https://github.com/shesha-io/shesha-framework/commit/38978b40ac91e6a1ed6fc7a460e3093c28d5e7ae) in response to your earlier review comment ("the function handles a data type that is not declared in the project… it must be explicitly declared"). Before that, normalizeConfigurableItemIdentifier took unknown and did the same shape sniffing untyped.
> 
> "The normalizeConfigurableItemIdentifier function handles a data type that is not declared in the project. If this type genuinely needs to be handled, it must be explicitly declared, and the corresponding calls must also be present. Otherwise, the code is both pointless and fragile…"

Hi @James-Baloyi. Correct, the previous revision of the proposed changes handled a data structure that wasn't defined as a type in the code, which is why I included the caveat "If this type genuinely needs to be handled." However, I don't see any actual usage of this functionality (I mean this type of arguments), so it appears to be unnecessary.
This is core functionality that accepts only two well-defined inputs: a raw ID (GUID as a string) and an object containing a mandatory item name and module name. It should not be cluttered with type-conversion logic—the argument type itself is a strict requirement here.

James-Baloyi:
Morning @IvanIlyichev  

I deleted PersistedConfigurableItemRawId, PersistedConfigurableItemFullName, PersistedConfigurableItemIdentifier, the canonicalModule helper and normalizeConfigurableItemIdentifier entirely. The file is back to the two guards plus configurableItemIdentifierToString and the bug fixes are still intact.

IvanIlyichev (approved):

Issue #5012, opened 2026-06-10
Form Designer: Canvas size and zoom control improvements


AshleighMab:
**Description**
The Form Designer canvas and zoom controls in Configuration Studio have several related usability issues that make it difficult to work effectively on form layouts. These are grouped here as they all relate to the same toolbar and canvas area.

**5a - Canvas defaults to 1920px instead of filling available space**
When a user opens a form in the Form Designer, the canvas is rendered at a fixed 1920px width. This does not account for the space actually available between the Builder Components panel on the left and the Properties panel on the right. The canvas should instead fill the available viewport width dynamically using dvw/dvh units, so it correctly fits the working area regardless of screen resolution or panel widths.

**5b - Default zoom is "Auto" and components are too small to work with**
The canvas defaults to "Auto" zoom mode, which at a 1920px canvas produces a zoom level of approximately 47% on typical screens. At this scale, form components are too small to interact with meaningfully. The default zoom should be set to 80% so that the canvas is immediately usable when a form is opened.

**5c - Zoom increments in 1% steps - should use predefined levels**
The + and - zoom buttons currently increment and decrement zoom by 1% per click. This makes it impractical to navigate between useful zoom levels. The zoom control should instead step through a predefined set of meaningful levels, for example: 25%, 50%, 75%, 80%, 100%, 125%, 150%, 200%.

**5d - Zoom percentage between + and - buttons should be manually editable**
The zoom percentage indicator displayed between the + and - buttons is currently read-only. It should be replaced with an editable input field so users can type a specific zoom value directly, without having to click through incremental steps.

**5e - In Auto mode, + and - buttons are disabled and should exit Auto mode**
When the zoom mode is set to "Auto", the + and - zoom buttons are non-interactive (read-only). Instead of being disabled, clicking either button while in Auto mode should automatically switch the zoom mode to Manual and apply the next stepped zoom level up or down from the current auto-computed value. This gives users a smooth path from auto zoom to manual control without having to explicitly change the mode selector first.

**5f - Resolution dropdown "Default" option should be renamed to "Canvas"**
The first option in the resolution/breakpoint dropdown is currently labelled "Default". This label is ambiguous — it does not convey what the option actually does. It should be renamed to "Canvas" to clearly indicate that selecting it uses the full available canvas width, as opposed to a specific device breakpoint.

**5g - Add 1920x1080 as a preset option in the resolution dropdown**
[...]

PR #5317, opened 2026-08-20
Implement auto width feature for canvas and sidebar actions

micanipho:
Fix: SidebarContainer now measures the pane between the sidebars with a ResizeObserver and lays the canvas out at availableWidth / zoom, so after CSS zoom is
  applied it renders exactly pane-width. Zooming out widens the layout (components re-wrap) while each component keeps the scale the user picked. A new autoWidth
  flag drives the mode; picking any device preset turns it off and pins the width.

  Also in here:
  - Auto zoom is disabled in Canvas mode — it exists to fit a fixed-width canvas, and reacting to a width derived from the zoom would feed back on itself.
  - Default zoom 80% → 75%; zoom levels regularised to 25% steps.
  - localStorage keys bumped to :v2 — otherwise a zoom persisted by the old version (including odd auto-zoom values like 47%) would override the new default. This
    resets saved width/zoom for existing users.
  - Fixes: clearing the zoom input jumped the canvas to 10% (parser returned 0 instead of NaN); the dropdown showed "Canvas" for any device preset whose width
    happened to match the screen; clampZoom was duplicated in two files and didn't handle non-finite input.

## Summary by CodeRabbit

* **New Features**
  * Added responsive canvas sizing that adapts to the available workspace.
  * Added automatic and fixed-width options, including percentage-based widths up to 100%.
  * Canvas width and zoom preferences are preserved between sessions.

* **Improvements**
  * Disabled auto-zoom and manual zoom controls while automatic sizing is active.
  * Updated zoom presets, including a 75% default and expanded range.
  * Improved zoom validation, pinch-zoom behavior, and content expansion.

* **Bug Fixes**
  * Prevented horizontal overflow when the canvas fits the workspace.
  * Restored the previous zoom value after clearing the zoom field.

Related to: #5012


czwe-01 (changes requested):
Screenshot 2026-08-28 at 08 38 05

@micanipho  please address the following issues:
1. Canvas mode must resolve to a real device for styling

The "Canvas" preset is a sizing mode, not a device. Styling only supports desktop / mobile / tablet, so while Canvas is active the canvas must still resolve to one of those three and apply that device's styles.

On initialization, don't leave the device unresolved or defaulted just because the width preset is Canvas. Initialize it from the last configured screen so the form opens with the styles that were last in effect.
While in Canvas mode, derive the device from the measured pane width, and keep it updated as the pane resizes.

2. Content after a 100vh component escapes the canvas instead of scrolling

Repro is in the screenshot: Container1 is set to height: 100vh and fits the canvas exactly, which is correct. But everything after it — Gender, Dropdown2, Entity Picker1, Date field1 — renders below the canvas boundary, outside the visible canvas area, rather than being reachable by scrolling.

PR #5449, opened 2026-09-03
Update configurations for forms on Shesha module

micanipho:
- otp-verification
 - registration


IvanIlyichev (changes requested):
Hi @micanipho. Attached package contains forms from  internal module `Boxfusion.SheshaFunctionalTests.Common`. It shouldn't be a part of framework. It these forms are for internal testingonly they should be attached to `Boxfusion.SheshaFunctionalTests.Common`, it it's framework part they should be created under Shesha module.
PS: next time please assign me as a reviewer, not as main assignee

IvanIlyichev (approved):

PR #5531, opened 2026-09-16
fix(nuget): update local-packages path in nuget.config

micanipho:


IvanIlyichev (approved):
Hi @micanipho. This file can be safely removed.

